PR #3, opened 2017-07-18
add kronos deployment artifact

surajssd:


vpavlin:
Hi Suraj, 

we cannot use this for production or staging deployment - we need to be able to parameterise the IMAGE_TAG in so for version tracking. 

Also my knowledge of kedge is minimal so we will need to talk before we create a dependency on it in our OpenShift.io prod environment

app.yaml
```diff
@@ -0,0 +1,42 @@
...
+  resources:
+    requests:
+      memory: 512Mi
+      cpu: 4
+  command:
+  - /bin/entrypoint.sh
+services:
+- ports:
+  - port: 6006
+configMaps:
```
vpavlin: @surajssd How is this used in container without referencing it in env section?:-)

surajssd: It is referenced in the [`envFrom`](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/pull/3/files#diff-bb68be076839a38efa2ebc731942536aR6) section!

surajssd:
> we cannot use this for production or staging deployment - we need to be able to parameterise the IMAGE_TAG in so for version tracking.

the image I created so that we can try this out atleast, and am sure this image should come from some form of build pipeline!

surajssd:
> Also my knowledge of kedge is minimal so we will need to talk before we create a dependency on it in our OpenShift.io prod environment

Sure

vpavlin:
I am not concerned about the image itself, I am concerned about ability of Kedge to process templates

msrb:
on an unrelated note, where can I find more information about Kedge?

vpavlin:
@msrb https://github.com/kedgeproject/kedge

surajssd:
@vpavlin @msrb updated the PR to not use kedge and only use the openshift artifacts

vpavlin:
@surajssd If you don't mind I will request more changes from you to get is as close to production as possible, unless it's so far from production that guys happy to merge it like this and iterate later. 

The requirements are:
* Move the yaml(s) into `openshift/` subdirectory (just for best practices, no real need for it..)
* Change the cpu requests to 3 or <, I'd suggest put there a parameter with default value 1, we can override it in prod
* Make it a template
* Add `IMAGE_TAG` parameter
* Use `registry.devshift.net/...` for image - once @msrb merges his Jenkinsfile image is going to get there anyways..
* Put env vars into configMap
* Put the configMap to a separate file
* Rename the items in configMap so that they do not reference Tuhin
* Add secret yaml with some dummy values

@msrb @TuhinSharma @surajssd Your call, guys:) Feel free to merge, but please keep these somewhere as an issue or something..

msrb:
Thanks @surajssd! This is definitely a good start. Like Vasek said, there is still some work left, but I am OK with merging this and then more people can iterate on it.

TuhinSharma:
Thanks @msrb and @vpavlin for the feedback. @surajssd Can you please create an issue and keep track of the points suggested by @vpavlin ?. I am merging it for now.

PR #9, opened 2017-07-25
Update template

msrb:


msrb:
@tuhinsharma PTAL.

tuhinsharma:
@msrb LGTM.

PR #10, opened 2017-07-28
Improve Intent Hierarchy of Gnosis Ref. Architecture

tuhinsharma:
Previously, the Reference architecture had only 1 level of intent. Now 2nd level of intent is introduced. The logic for activating the intent nodes in the Kronos PGM is also modified. The default recommendation is fixed and the recommendation is improved for arbitrary stacks as well.


tuhinsharma:
@miteshvp @msrb Please look into this. :-)

PR #13, opened 2017-08-01
introduce kronos service for pypi/maven

miteshvp:
@msrb please see if this is sufficient. I will make necessary changes in worker side then


msrb:
LGTM, thanks 😉

PR #24, opened 2017-08-02
Lowering number of workers to 2

msrb:
This reverts commit 40d60a11ef29f4700b39f7d732b621373aba4522.

### DIFF
--- a/analytics_platform/kronos/scripts/entrypoint.sh
+++ b/analytics_platform/kronos/scripts/entrypoint.sh
@@ -6,7 +6,7 @@
 
 zip -r /tmp/training.zip /analytics_platform /util
 
-gunicorn --pythonpath / -b 0.0.0.0:$SERVICE_PORT --workers=10 -t $SERVICE_TIMEOUT rest_api:app
+gunicorn --pythonpath / -b 0.0.0.0:$SERVICE_PORT --workers=2 -t $SERVICE_TIMEOUT rest_api:app
 
 # --------------------------------------------------------------------------------------------------
 # to make the container alive for indefinite time


msrb:
Trying to run the container with more than 1 gunicorn worker means that code in https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/deployment/rest_api.py#L26-L32 gets executed multiple times (10 workers=10 processes).

And it seems like the service never gets ready because of that. Therefore reverting for now.

msrb:
@tuhinsharma I can see that pypi route is already configured with timeout of 300 seconds.

msrb:
Actually the service seems to be running now, marking as wip and checking... sorry for the fuss.

samuzzal-choudhury:
@msrb IIUC, the gateway timeout is caused by a timeout configuration at the openshift level and nothing to do with the application.

msrb:
The containers have been restarted 20+ times, lowering number of workers to 2.

msrb:
And merging to see if it helps :)

msrb:
@samuzzal-choudhury 

> IIUC, the gateway timeout is caused by a timeout configuration at the openshift level and nothing to do with the application.

That sounds about right. The timeout on the route is now set to 900 seconds. But the container is getting restarted every ~3 minutes when request is being processed. It seems to be caused by readiness/liveness probes. Those probes try to ping the Kronos API, but the system can currently handle only 2 concurrent requests at any given time.

```bash
$ time curl -vvv --max-time 900 -XPOST -H 'Content-Type: application/json' -d '[{"ecosystem": "pypi","comp_package_count_threshold": 2,"alt_package_count_threshold": 4,"outlier_probability_threshold": 0.8,"unknown_packages_ratio_threshold": 0.3,"package_list": ["requests","scikit-learn","coverage","cycler","numpy","mock","nose","scipy","matplotlib","nltk","pandas"]}]'  http://bayesian-kronos-pypi-bayesian-preview.b6ff.rh-idev.openshiftapps.com/api/v1/schemas/kronos_scoring
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 52.207.133.22...
* TCP_NODELAY set
* Connected to bayesian-kronos-pypi-bayesian-preview.b6ff.rh-idev.openshiftapps.com (52.207.133.22) port 80 (#0)
> POST /api/v1/schemas/kronos_scoring HTTP/1.1
> Host: bayesian-kronos-pypi-bayesian-preview.b6ff.rh-idev.openshiftapps.com
> User-Agent: curl/7.51.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 290
> 
* upload completely sent off: 290 out of 290 bytes
* HTTP 1.0, assume close after body
< HTTP/1.0 502 Bad Gateway
< Cache-Control: no-cache
< Connection: close
< Content-Type: text/html
< 
<html><body><h1>502 Bad Gateway</h1>
The server returned an invalid or incomplete response.
</body></html>
* Curl_http_done: called premature == 0
* Closing connection 0

real	3m11.671s
user	0m0.015s
sys	0m0.004s
```

PR #27, opened 2017-08-03
Include Topics in recommendation response

tuhinsharma:
This PR has the following changes:-
1. Topics are included in the alternate package, companion package and outlier packages.
2. Topics for observed packages are included as well.
3. Number of outlier packages are made a maximum of 2 given they have outlier probability greater than 0.9 (temporary fix)
4. Similar package calculation logic is improved.
5. Due to the addition of features some unit tests are modified.


tuhinsharma:
@msrb @miteshvp @krishnapaparaju need your help.

PR #33, opened 2017-08-07
[jenkins] Fix rollbacks

msrb:


sara-02:
@msrb Do we still need this?
@miteshvp Can you please review.

PR #43, opened 2017-08-28
Use just 1 CPU in staging

msrb:


msrb:
rh-idev has only 5 nodes, 3.5 CPU cores each. And fabric8-analytics is not the only project running there.

sara-02:
@msrb Thanks.

PR #45, opened 2017-09-12
Python3 compatible code

rootAvish:
Made some print statements that were still written in Python2 style Python3 compatible, were the only ones that wouldn't work.

### DIFF
--- a/analytics_platform/kronos/deployment/rest_api.py
+++ b/analytics_platform/kronos/deployment/rest_api.py
@@ -8,21 +8,22 @@ from flask import Flask, request, g
 from flask_cors import CORS
 
 from analytics_platform.kronos.deployment.submit_training_job import submit_job
-from analytics_platform.kronos.gnosis.src.gnosis_constants import *
+import analytics_platform.kronos.gnosis.src.gnosis_constants as gnosis_constants
 from analytics_platform.kronos.pgm.src.offline_training import load_eco_to_kronos_dependency_dict_s3
 from analytics_platform.kronos.src.config import (
     AWS_BUCKET_NAME, KRONOS_MODEL_PATH, KRONOS_SCORING_REGION)
-from analytics_platform.kronos.src.kronos_online_scoring import *
+from analytics_platform.kronos.src.kronos_online_scoring import \
+    load_user_eco_to_kronos_model_dict_s3, score_eco_user_package_dict
 from analytics_platform.kronos.src.recommendation_validator import RecommendationValidator
 from tagging_platform.helles.deployment.submit_npm_tagging_job import submit_tagging_job
 from tagging_platform.helles.npm_tagger.get_descriptions_from_s3 import run as \
     run_description_collection
 from tagging_platform.helles.npm_tagger.get_version_info_for_missing_packages import run_job as \
     run_missing_package_version_collection_job
-
+import six
 
 if sys.version_info.major == 2:
-    reload(sys)
+    six.reload(sys)
     sys.setdefaultencoding('UTF8')
 
 logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
@@ -91,10 +92,13 @@ def train_and_save_kronos():
 
     input_json = request.get_json()
     training_data_url = input_json.get("training_data_url")
-    fp_min_support_count = input_json.get(FP_MIN_SUPPORT_COUNT_NAME, FP_MIN_SUPPORT_COUNT_VALUE)
-    fp_intent_topic_count_threshold = input_json.get(FP_INTENT_TOPIC_COUNT_THRESHOLD_NAME,
-                                                     FP_INTENT_TOPIC_COUNT_THRESHOLD_VALUE)
-    fp_num_partition = input_json.get(FP_NUM_PARTITION_NAME, FP_NUM_PARTITION_VALUE)
+    fp_min_support_count = input_json.get(gnosis_constants.FP_MIN_SUPPORT_COUNT_NAME,
+                                          gnosis_constants.FP_MIN_SUPPORT_COUNT_VALUE)
+    fp_intent_topic_count_threshold = input_json.get(
+        gnosis_constants.FP_INTENT_TOPIC_COUNT_THRESHOLD_NAME,
+        gnosis_constants.FP_INTENT_TOPIC_COUNT_THRESHOLD_VALUE)
+    fp_num_partition = input_json.get(gnosis_constants.FP_NUM_PARTITION_NAME,
+                                      gnosis_constants.FP_NUM_PARTITION_VALUE)
 
     response = submit_job(input_bootstrap_file='/bootstrap_action.sh',
                           input_src_code_file='/tmp/training.zip',
@@ -131,7 +135,6 @@ def predict_and_score():
 @app.route('/api/v2/npm_tagging', methods=['POST'])
 def tag_npm_packages_textrank():
     input_json = request.get_json()
-    # if 'package_name' in input_json:
     response = submit_tagging_job(input_bootstrap_file='/helles_bootstrap_action.sh',
                                   input_src_code_file='/tmp/tagging.zip',
                                   package_name=input_json.get('package_name', ''),
--- a/analytics_platform/kronos/deployment/rest_api.py
+++ b/analytics_platform/kronos/deployment/rest_api.py
@@ -22,8 +22,8 @@ from tagging_platform.helles.npm_tagger.get_version_info_for_missing_packages im
     run_missing_package_version_collection_job
 import six
 
-if sys.version_info.major == 2:
-    six.reload(sys)
+if six.PY2:
+    reload(sys)
     sys.setdefaultencoding('UTF8')
 
 logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
--- a/analytics_platform/kronos/deployment/rest_api.py
+++ b/analytics_platform/kronos/deployment/rest_api.py
@@ -20,11 +20,7 @@ from tagging_platform.helles.npm_tagger.get_descriptions_from_s3 import run as \
     run_description_collection
 from tagging_platform.helles.npm_tagger.get_version_info_for_missing_packages import run_job as \
     run_missing_package_version_collection_job
-import six
 
-if six.PY2:
-    reload(sys)
-    sys.setdefaultencoding('UTF8')
 
 logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
 
--- a/analytics_platform/kronos/deployment/submit_training_job.py
+++ b/analytics_platform/kronos/deployment/submit_training_job.py
@@ -10,10 +10,15 @@
 # 1. Improve error handling
 # ----------------------------------------------------------------------------------------------------------------------
 import sys
+import logging
 import boto3
 from time import gmtime, strftime
+import daiquiri
+
 from analytics_platform.kronos.src import config
-from util.logging.project_logger import logger as _logger
+
+daiquiri.setup(level=daiquiri.setup(level=logging.INFO))
+_logger = daiquiri.getLogger(__name__)
 
 COMPONENT_PREFIX = "kronos"
 
--- a/analytics_platform/kronos/deployment/submit_training_job.py
+++ b/analytics_platform/kronos/deployment/submit_training_job.py
@@ -17,7 +17,7 @@ import daiquiri
 
 from analytics_platform.kronos.src import config
 
-daiquiri.setup(level=daiquiri.setup(level=logging.INFO))
+daiquiri.setup(level=logging.INFO)
 _logger = daiquiri.getLogger(__name__)
 
 COMPONENT_PREFIX = "kronos"
--- a/analytics_platform/kronos/deployment/submit_training_job.py
+++ b/analytics_platform/kronos/deployment/submit_training_job.py
@@ -13,6 +13,7 @@ import sys
 import boto3
 from time import gmtime, strftime
 from analytics_platform.kronos.src import config
+from util.logging.project_logger import logger as _logger
 
 COMPONENT_PREFIX = "kronos"
 
@@ -36,17 +37,17 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
                                                str_cur_time)
     s3_log_uri = 's3://{bucket}/{key}'.format(bucket=s3_log_bucket, key=s3_log_key)
 
-    print("Uploading the bootstrap action to AWS S3 URI {} ...".format(s3_bootstrap_uri))
+    _logger.debug("Uploading the bootstrap action to AWS S3 URI {} ...".format(s3_bootstrap_uri))
     # Note: This overwrites if file already exists
     s3_client = boto3.client('s3',
                              aws_access_key_id=config.AWS_S3_ACCESS_KEY_ID,
                              aws_secret_access_key=config.AWS_S3_SECRET_ACCESS_KEY)
     s3_client.upload_file(input_bootstrap_file, s3_bucket, s3_bootstrap_key)
 
-    print("Uploading the src code to AWS S3 URI {} ...".format(s3_uri))
+    _logger.debug("Uploading the src code to AWS S3 URI {} ...".format(s3_uri))
     s3_client.upload_file(input_src_code_file, s3_bucket, s3_key)
 
-    print("Starting spark emr cluster and submitting the jobs ...")
+    _logger.debug("Starting spark emr cluster and submitting the jobs ...")
     emr_client = boto3.client('emr',
                               aws_access_key_id=config.AWS_S3_ACCESS_KEY_ID,
                               aws_secret_access_key=config.AWS_S3_SECRET_ACCESS_KEY,
@@ -168,6 +169,6 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
         output['training_job_id'] = "Error"
         output['status'] = 'Error'
         output['status_description'] = "Error! The job/cluster could not be created!"
-        print(response)
+        _logger.debug(response)
 
     return output
--- a/analytics_platform/kronos/deployment/submit_training_job.py
+++ b/analytics_platform/kronos/deployment/submit_training_job.py
@@ -9,7 +9,6 @@
 # TODO:
 # 1. Improve error handling
 # ----------------------------------------------------------------------------------------------------------------------
-import sys
 import logging
 import boto3
 from time import gmtime, strftime
--- a/analytics_platform/kronos/deployment/submit_training_job.py
+++ b/analytics_platform/kronos/deployment/submit_training_job.py
@@ -36,17 +36,17 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
                                                str_cur_time)
     s3_log_uri = 's3://{bucket}/{key}'.format(bucket=s3_log_bucket, key=s3_log_key)
 
-    print "Uploading the bootstrap action to AWS S3 URI " + s3_bootstrap_uri + " ..."
+    print("Uploading the bootstrap action to AWS S3 URI {} ...".format(s3_bootstrap_uri))
     # Note: This overwrites if file already exists
     s3_client = boto3.client('s3',
                              aws_access_key_id=config.AWS_S3_ACCESS_KEY_ID,
                              aws_secret_access_key=config.AWS_S3_SECRET_ACCESS_KEY)
     s3_client.upload_file(input_bootstrap_file, s3_bucket, s3_bootstrap_key)
 
-    print "Uploading the src code to AWS S3 URI " + s3_uri + " ..."
+    print("Uploading the src code to AWS S3 URI {} ...".format(s3_uri))
     s3_client.upload_file(input_src_code_file, s3_bucket, s3_key)
 
-    print "Starting spark emr cluster and submitting the jobs ..."
+    print("Starting spark emr cluster and submitting the jobs ...")
     emr_client = boto3.client('emr',
                               aws_access_key_id=config.AWS_S3_ACCESS_KEY_ID,
                               aws_secret_access_key=config.AWS_S3_SECRET_ACCESS_KEY,
@@ -168,6 +168,6 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
         output['training_job_id'] = "Error"
         output['status'] = 'Error'
         output['status_description'] = "Error! The job/cluster could not be created!"
-        print response
+        print(response)
 
     return output
--- a/analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
+++ b/analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
@@ -237,7 +237,7 @@ class GnosisReferenceArchitecture(AbstractGnosis):
                 item_list, key=lambda x: x[1], reverse=False)
             topic_num_to_itemset_dict[key] = sorted_item_list
 
-        item_dict = {z: 10 for z in package_list}
+        item_dict = {z: gnosis_constants.FP_TAG_INTENT_LIMIT for z in package_list}
         intent_dict = dict()
         for key_value in topic_num_to_itemset_dict:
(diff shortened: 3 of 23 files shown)


miteshvp:
[test]

analytics_platform/kronos/deployment/submit_training_job.py
```diff
@@ -35,17 +35,17 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
     s3_log_key = '{}_{}_spark_emr_log_'.format(config.DEPLOYMENT_PREFIX, COMPONENT_PREFIX, str_cur_time)
     s3_log_uri = 's3://{bucket}/{key}'.format(bucket=s3_log_bucket, key=s3_log_key)
 
-    print "Uploading the bootstrap action to AWS S3 URI " + s3_bootstrap_uri + " ..."
+    print("Uploading the bootstrap action to AWS S3 URI " + s3_bootstrap_uri + " ...")
```
pkajaba: Hey @rootAvish, maybe you can take a look into `format` method instead of concating string :-):
`print("Uploading the bootstrap action to AWS S3 URI {} ...".format(s3_bootstrap_uri))`

analytics_platform/kronos/deployment/submit_training_job.py
```diff
@@ -35,17 +35,17 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
...
     # Note: This overwrites if file already exists
     s3_client = boto3.client('s3',
                              aws_access_key_id=config.AWS_S3_ACCESS_KEY_ID,
                              aws_secret_access_key=config.AWS_S3_SECRET_ACCESS_KEY)
     s3_client.upload_file(input_bootstrap_file, s3_bucket, s3_bootstrap_key)
 
-    print "Uploading the src code to AWS S3 URI " + s3_uri + " ..."
+    print("Uploading the src code to AWS S3 URI " + s3_uri + " ...")
```
pkajaba: same here

util/http_client/request.py
```diff
@@ -7,7 +7,7 @@ def post(url, json_data):
 
     # TODO: check for error and raise exception
     if response.status_code != 200:
-        print ("ERROR %d: %s") % (response.status_code, response.reason)
+        print(("ERROR %d: %s") % (response.status_code, response.reason))
```
pkajaba: again :-)

analytics_platform/kronos/deployment/submit_training_job.py
rootAvish: Yeah that's a good point, will make the change.

rootAvish:
[test]

rootAvish:
[test]

rootAvish:
Both training and testing work with Python 3, did the changes requested by @pkajaba .

analytics_platform/kronos/gnosis/src/offline_training.py
```diff
@@ -4,7 +4,7 @@
 from analytics_platform.kronos.gnosis.src.gnosis_package_topic_model import GnosisPackageTopicModel
 from analytics_platform.kronos.gnosis.src.gnosis_ref_arch import GnosisReferenceArchitecture
 from analytics_platform.kronos.src import config
-from gnosis_constants import *
```
pkajaba: I know that star import was here before, but you should change it in this PR since there no good reason to have it in code. Better way to do it is just `import analytics_platform.kronos.gnosis.src.gnosis_constants` and use it in code. 

I know that it will be unnecessarily long, but the problem is project structure since it does not follow standards.

analytics_platform/kronos/pgm/test/test_pgm_pomegranate.py
```diff
@@ -57,7 +57,7 @@ def test_score_eco_user_package_dict(self):
 
         eco_to_kronos_dependency_dict = get_eco_to_kronos_dependency_dict(data_store=input_data_store,
                                                                           folderpath="data_kronos_dependency")
-        print eco_to_kronos_dependency_dict
+        print(eco_to_kronos_dependency_dict)
```
pkajaba: I wonder why do you need this print in tests. Can you elaborate on this?

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
```diff
@@ -70,11 +70,11 @@ def score(self, evidence_dict_list):
         pgm_model_kronos = self.model
         n = len(evidence_dict_list)
         n_jobs = n
-        result_array = reduce(list.__add__, Parallel(n_jobs=n_jobs)(
+        result_array = functools.reduce(list.__add__, Parallel(n_jobs=n_jobs)(
             delayed(parallel_predict)(evidence_dict_list[int(n / n_jobs * i):int(n / n_jobs * (i + 1))]) for i in
-            range(n_jobs)))
+            list(range(n_jobs))))
```
pkajaba: do you really need to convert the range to list? IMHO, does not make a lot of sense.

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
```diff
@@ -70,11 +70,11 @@ def score(self, evidence_dict_list):
         pgm_model_kronos = self.model
         n = len(evidence_dict_list)
         n_jobs = n
```
pkajaba: I know that this is from previous code, but I don't see any good reason to have `n_jobs` variable here. I keep rereading it, but still no meaning coming to me.

analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
```diff
@@ -226,7 +226,7 @@ def _generate_intent_component_class_dict_fp_growth(cls, model, min_intent_topic
                     if item in item_dict:
                         item_dict[item] -= 1
                 keys_to_remove = list()
-                for key, value in item_dict.iteritems():
+                for key, value in list(item_dict.items()):
```
pkajaba: no need to have a list here.

analytics_platform/kronos/deployment/submit_training_job.py
```diff
@@ -35,17 +35,17 @@ def submit_job(input_bootstrap_file, input_src_code_file, training_data_url, fp_
     s3_log_key = '{}_{}_spark_emr_log_'.format(config.DEPLOYMENT_PREFIX, COMPONENT_PREFIX, str_cur_time)
     s3_log_uri = 's3://{bucket}/{key}'.format(bucket=s3_log_bucket, key=s3_log_key)
 
-    print "Uploading the bootstrap action to AWS S3 URI " + s3_bootstrap_uri + " ..."
+    print("Uploading the bootstrap action to AWS S3 URI {} ...".format(s3_bootstrap_uri))
```
pkajaba: I am sorry but I have just realized that in our `definition of done` there is a goal not to use `print` statements. You should set up logger instead of this. This library is really easy to use: https://github.com/jd/daiquiri

pkajaba:
A lot of comments here are about code which was committed previously but if you are familiar with the code it would be great if you rework my comments. You can do it another PR or in this one, I leave it up to you.

analytics_platform/kronos/deployment/submit_training_job.py
rootAvish: Do we use it throughout the rest of our projects as well? If not I guess it would be good to use the same library everywhere.

pkajaba: it's not unified yet.

rootAvish:
[test]

pkajaba:
@rootAvish Can you rebase this one?

rootAvish:
@pkajaba right, truth be told I didn't get around to this PR, agree with the review comments, will try and fix all this stuff over the weekend.

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
rootAvish: On the other hand, we can just remove n 🙂

analytics_platform/kronos/deployment/submit_training_job.py
rootAvish: I'll log everything at a "debug" level for now, then make a separate PR for what the appropriate log level should be everywhere.

analytics_platform/kronos/gnosis/src/offline_training.py
rootAvish: I did that change, but why not instead of a star import, import all the relevant constants and use them?

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
rootAvish: The entire thing was written in a really weird way, turns out we can make it much simpler.

analytics_platform/kronos/pgm/test/test_pgm_pomegranate.py
rootAvish: I've removed this in a different PR, it was a leftover according to me as well.

rootAvish:
@pkajaba I guess we can merge now 🙂

analytics_platform/kronos/gnosis/src/offline_training.py
fridex: > I did that change, but why not instead of a star import, import all the relevant constants and use them?

Not a good idea in general, see for example https://stackoverflow.com/questions/2386714/why-is-import-bad

rootAvish: @fridex I meant instead of `from gnosis_constants import *` we do - `from gnosis_constants import GNOSIS_RA_OUTPUT_PATH, GNOSIS_RA_OUTPUT_PATH, GNOSIS_PTM_OUTPUT_PATH`, and not what we did in where we have to prefix `analytics_platform.kronos.gnosis.src.gnosis_constants` before everything. What benefit does the third approach offer over the second?

rootAvish: The question you linked only covers why "*" imports are bad.

fridex: > @fridex I meant instead of from gnosis_constants import * we do - from gnosis_constants import GNOSIS_RA_OUTPUT_PATH, GNOSIS_RA_OUTPUT_PATH, GNOSIS_PTM_OUTPUT_PATH, and not what we did in where we have to prefix analytics_platform.kronos.gnosis.src.gnosis_constants before everything. What benefit does the third approach offer over the second?

The import seems really unnecessarily long (refactor sources?). Also you can import module under different name space using `import .. as ..` (we do it already in various places, see [example](https://github.com/fabric8-analytics/fabric8-analytics-jobs/blob/master/f8a_jobs/api_v1.py#L20)). As stated in the linked post on stack overflow, star imports are really not a good idea - maybe it seems harmless now but as software evolves, it can introduce a lot of harm. If you try to google this, you will find a lot of articles that explains why it is not good. Take a look at [this article](https://pythonconquerstheuniverse.wordpress.com/2011/03/28/why-import-star-is-a-bad-idea/) as a first-google-result-example.

fridex: Maybe I missed your question here. If you are asking why there should  be used full path to module on import - if you get this python script and run it, it will not work. The reason behind it is that there is adjusted python path in the [enttypoint](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/scripts/entrypoint.sh#L9). It relies on this hack, it is confusing (as one is not sure from where one imports things without knowing this import path change) and all static code analysers will simply report errors.

rootAvish: I'm sold on star imports being bad, I don't use them personally, this was just old code that was already written this way.

> The import seems really unnecessarily long (refactor sources?)

Yeah, but prefixing module name to constants everywhere seemed too long to me 🙂 

>  you are asking why there should be used full path to module on import 

No, but nevermind, I think I might've interpreted Pavel's suggestion in the wrong way.

pkajaba:
[test]

rootAvish:
@pkajaba Do you want me to squash some of the initial commits here? This PR ended up covering much more ground than those first few commits sought to 🙂

pkajaba:
@rootAvish That would be great.

rootAvish:
I guess this should be good.

analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
```diff
@@ -8,8 +8,8 @@
 
 from analytics_platform.kronos.gnosis.src.abstract_gnosis import AbstractGnosis
 from analytics_platform.kronos.gnosis.src.gnosis_package_topic_model import GnosisPackageTopicModel
-from gnosis_constants import *
-from gnosis_util import *
+from analytics_platform.kronos.gnosis.src.gnosis_constants import *
```
pkajaba: as we talked before we don't want to do star imports :-(

analytics_platform/kronos/gnosis/src/offline_training.py
```diff
@@ -19,7 +19,9 @@ def train_and_save_gnosis_ref_arch(input_data_store, output_data_store, addition
         min_support_count=fp_min_support_count,
         min_intent_topic_count=fp_intent_topic_count_threshold,
         fp_num_partition=fp_num_partition)
-    gnosis_ref_arch_obj.save(output_data_store, additional_path + GNOSIS_RA_OUTPUT_PATH)
+    gnosis_ref_arch_obj.save(
```
pkajaba: can you use standard library call to concatenate strings here?

analytics_platform/kronos/gnosis/src/offline_training.py
```diff
@@ -60,11 +62,13 @@ def generate_and_save_gnosis_package_topic_model(input_data_store, output_data_s
 
     gnosis_package_topic_model_obj = GnosisPackageTopicModel.curate(
         data_store=input_data_store,
-        filename=additional_path + GNOSIS_PTM_INPUT_PATH,
+        filename=additional_path +
```
pkajaba: same thing applies here

util/logging/project_logger.py
```diff
@@ -0,0 +1,5 @@
+import daiquiri
+import logging
+
+daiquiri.setup(level=logging.DEBUG)
+logger = daiquiri.getLogger(__name__)
```
pkajaba: Please, can you move `getLogger` call to modules where you use it?  If you use it this way it might lead to confusing logs since project_logger would be used in every log output instead real module names.

analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
rootAvish: IDK why but these are everywhere.. I thought they had been used in just that one place.

analytics_platform/kronos/gnosis/src/offline_training.py
rootAvish: You mean using `os.path.join`?

analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
pkajaba: well, I am sorry about pointing out just now. However, it's really bad practice, since it can override standard libraries. This can be OK in some cases, but I am afraid here it's just sign of laziness.

analytics_platform/kronos/gnosis/src/offline_training.py
pkajaba: yes.

rootAvish:
@pkajba, please take a look, addressed everything and got rid of every last * import. There's a whole lot of housekeeping in this PR in addition to the Python3 changes that makes QOL in working with Kronos much easier, so we need to merge this before it gets any bigger. We can carry on with related stuff in other, smaller PRs. The real problem here is that the first draft of this code was never reviewed before this came to me and to be quite honest there's a lot of mess here.

On a side note if there's questionable stuff like the constants imported via utils as in #117, would appreciate it if you don't let it pass, only creates more technical debt which created a lot of additional work in this PR itself. Also the imports in it were not Python3 compatible 🙁 , we shouldn't be allowing those in either.

pkajaba:
> On a side note if there's questionable stuff like the constants imported via utils as in #117, would appreciate it if you don't let it pass, only creates more technical debt which created a lot of additional work in this PR itself

I can't agree more, but @sara-02 created an issue about it here: https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/issues/129
I won't happen again, but I did not to be too strict since there were a lot of changes in that PR anyway.

> Also the imports in it were not Python3 compatible :slightly_frowning_face: , we shouldn't be allowing those in either.

Yeah, but in that PR whole repo was not Python3 compatible. I suggest prioritizing work.

rootAvish:
> On top of that, you should install Python3 in Dockerfile instead of Python2 to avoid this confusion.

I wanted to open a separate PR for that to make sure that we don't blame a dependency problem on the code so it's not a part of this.

pkajaba:
OK, It's your call. 

BTW, could you rebase it so you got rid of merge commits?

rootAvish:
I'm not sure how the latest version of pomegranate will hold up against Python3, there's work going on already to migrate this codebase to it since it has added out of the box support for parallelism, @abs51295 is on it.

Given the choice of latest pomegranate on a Python2 container versus current pomegranate on a Python3 container I would personally pick the former.

rootAvish:
@pkajaba #135 is the relevant commit, I think it should be merged only once @abs51295 has verified that the latest pomegranate works well with Python3.

analytics_platform/kronos/pgm/src/offline_training.py
```diff
@@ -1,8 +1,8 @@
 import sys
```
pkajaba: unused imports

util/data_store/local_filesystem.py
```diff
@@ -88,10 +88,10 @@ def read_pomegranate_model(self, filename):
     def byteify(cls, input):
         if isinstance(input, dict):
             return {LocalFileSystem.byteify(key): LocalFileSystem.byteify(value)
-                    for key, value in input.iteritems()}
+                    for key, value in input.items()}
         elif isinstance(input, list):
             return [LocalFileSystem.byteify(element) for element in input]
-        elif isinstance(input, unicode):
+        elif isinstance(input, str):
```
pkajaba: Don't you need to create if python2 here too? `str` is not same thing as `unicode` in Python2

analytics_platform/kronos/softnet/src/offline_training.py
```diff
@@ -1,19 +1,20 @@
 import sys
 import time
```
pkajaba: unused imports again

analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
```diff
@@ -236,7 +237,7 @@ def _generate_intent_component_class_dict_fp_growth(cls, model, min_intent_topic
                 item_list, key=lambda x: x[1], reverse=False)
             topic_num_to_itemset_dict[key] = sorted_item_list
 
-        item_dict = {z: 2 for z in package_list}
+        item_dict = {z: 10 for z in package_list}
```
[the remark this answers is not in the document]
rootAvish: It's not associated to this PR, came in the rebase, but it's the maximum number of intents a component class label can be a part of. We tried increasing the limit and we were able to go up to 10 at least with our currently limited data.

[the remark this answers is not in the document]
pkajaba: Make it as a parameter.

analytics_platform/kronos/deployment/rest_api.py
```diff
@@ -91,10 +92,13 @@ def train_and_save_kronos():
```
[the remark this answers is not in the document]
rootAvish: I created one #136, trying to fix it all in one go is leading to a situation where we're stuck with none of these changes and rising technical debt.

analytics_platform/kronos/deployment/submit_training_job.py
```diff
@@ -10,10 +10,16 @@
...
 # ----------------------------------------------------------------------------------------------------------------------
 import sys
+import logging
 import boto3
 from time import gmtime, strftime
+import daiquiri
+
 from analytics_platform.kronos.src import config
 
+daiquiri.setup(level=logging.INFO)
```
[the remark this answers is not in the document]
rootAvish: Yes, the EMR response from Boto at the INFO level is required if something goes wrong.

util/data_store/local_filesystem.py
rootAvish: This code is only used to run unit tests, they passed so I'm assuming it worked as expected.

analytics_platform/kronos/deployment/rest_api.py
```diff
@@ -8,20 +8,21 @@
...
 from tagging_platform.helles.deployment.submit_npm_tagging_job import submit_tagging_job
 from tagging_platform.helles.npm_tagger.get_descriptions_from_s3 import run as \
     run_description_collection
 from tagging_platform.helles.npm_tagger.get_version_info_for_missing_packages import run_job as \
     run_missing_package_version_collection_job
+import six
 
-
-if sys.version_info.major == 2:
+if six.PY2:
```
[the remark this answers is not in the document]
rootAvish: Makes sense. Thanks for the pointer.

analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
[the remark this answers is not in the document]
rootAvish: Sure.

analytics_platform/kronos/pgm/src/offline_training.py
rootAvish: Leftovers I guess, got rid of them.

util/data_store/local_filesystem.py
rootAvish: I still changed it so nobody else has the same confusion.

rootAvish: For Python3, it's okay to switch to the else condition since all strings are unicode anyway.

rootAvish:
@pkajaba I think I covered all your concerns.

rootAvish:
@pkajaba Ping. Sorry to rush you like this but this item would be on the back-burner again if we don't merge it in this week as there's other task items I'm required to take up. Are we good to merge?

analytics_platform/kronos/deployment/submit_training_job.py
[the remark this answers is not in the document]
fridex: > Yes, the EMR response from Boto at the INFO level is required if something goes wrong.

I would suggest to use log levels based on severity - if something goes wrong, it's not info, but rather warning or error so it is easily searchable in logs and not suppressed in the output.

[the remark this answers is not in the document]
rootAvish: We don't have the problem of deep logs since on EMR we have one log file per training job, we prefer to log everything because it also prints out all the data which we often need to check what went wrong.

Going wrong in our context is not restricted to exceptions, it's generally about our model not behaving how we want it to because there was overfitting or something.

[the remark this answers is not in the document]
fridex: > We don't have the problem of deep logs since on EMR we have one log file per training job, we prefer to log everything because it also prints out all the data which we often need to check what went wrong.

This was just my suggestion when I saw this discussion - do it how you consider to be useful for you. Note that if a system produces few gigabytes of logs per day (that is the case of fabric8-analytics), keeping track of severity of logged events across services will help you a lot with debugging/health checking/capturing system misbehaviour/...

> Going wrong in our context is not restricted to exceptions, it's generally about our model not behaving how we want it to because there was overfitting or something.

I'm not speaking about exceptions but rather expected and unexpected events that are logged based on severity.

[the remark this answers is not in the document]
rootAvish: Fair enough, I'll do a full review of the logging levels across Kronos and raise them wherever they're not required. I'll cover it outside this PR, added it to my housekeeping issue here: https://github.com/openshiftio/openshift.io/issues/1475

pkajaba:
Thanks for changes.

PR #53, opened 2017-09-18
Pass the filtered alternate list not original one

sara-02:
Pass the alternate list after filtering with respect to outliers.


miteshvp:
[test]

PR #60, opened 2017-09-21
[WIP]Restructure all tests under one test folder

sara-02:
Restructuring the test folder for better unit testing.


analytics_platform/kronos/gnosis/src/gnosis_ref_arch.py
```diff
@@ -1,3 +1,7 @@
+import sys
+sys.path.insert(1, '/usr/local/spark/python')
+sys.path.insert(2, '/usr/local/spark/python/lib/py4j-0.10.4-src.zip')
```
fridex: This looks suspicious. Why is it needed?

sara-02: Because the `test_kronos_gnosis_ref_arch` runs a spark job. But importing `pyspark` will give an error if this is not specified. In the IDEs we can set the spark path and tests will not fail. I faced an issue while running from command line.
One alternate that just came to my mind is to be able to specify the spark path in the `env`.
Otherwise this will be needed to properly invoke `pyspark`.

fridex: > Because the test_kronos_gnosis_ref_arch runs a spark job.

How do you invoke these tests?

sara-02: If you pull from this branch then the command is `python -m unittest discover tests -v`

fridex: I would rather suggest to create a script (let's say `run_test.sh`) that would setup environment and adjust PYTHONPATH rather there.

Also pytest has some nice features over unittest - we already use this framework elsewhere.

These are some recommendations to "future", you don't need to rework it now.

sara-02: @fridex Thanks, I will look into the script. The test cases were originally written using unittest, that is why did not change it.

sara-02:
@fridex There is still some work left in this, can we wait until tomorrow for merging. It is still a WIP :/

fridex:
> @fridex There is still some work left in this, can we wait until tomorrow for merging. It is still a WIP :/

Oh, sorry, missed WIP prefix. OK, I will revert this and you can continue.

PR #63, opened 2017-09-21
Moved pomegranate to requirements.txt

rootAvish:
It's kind of a nuisance when setting up this repository for development to manually look up the version of `pomegranate` in the `Dockerfile` and install it separately. Added it to requirements, didn't see a reason why it was being installed separately.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,6 @@ RUN yum install -y epel-release && \
 # --------------------------------------------------------------------------------------------------
 COPY ./analytics_platform/kronos/requirements.txt /
 RUN pip install -r /requirements.txt && rm /requirements.txt
-RUN pip install pomegranate==0.7.3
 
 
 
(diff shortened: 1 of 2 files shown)


rootAvish:
[test]

rootAvish:
@fridex we need to revert this, apparently the tests didn't run when this was merged and now it has broken master, does it not build automatically when something is merged to master? If it should then something's broken in the CI/CD.

fridex:
> @fridex we need to revert this, apparently the tests didn't run when this was merged and now it has broken master, does it not build automatically when something is merged to master? If it should then something's broken in the CI/CD.

There were some issues this morning, let's revert this one. Thanks for bringing this up.

tisnik:
@rootAvish @fridex could you please try to explain what's wrong with this PR? Does the pomegranate really needs to be installed as the last package with a separate command? if yes, please document it in requirements.txt. TYVM

rootAvish:
@tisnik It's some upstream bug in dependency resolution, pomegranate module does not build in the same command as numpy for some reason. Either we install pomegranate separately or we install numpy separately before installing requirements. It's documented here: https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis-base/blob/master/Dockerfile#L14.

I'm yet to check whether this issue persists with the newer versions. Will do that and report back.

PR #66, opened 2017-09-25
Changes to make the PGM training work in absence of tags for certain packages

rootAvish:
Introduces a fallback for training the PGM with packages
that do not have any topics associated with them. In these cases we
will tag the packages using descriptive parts of the package name
itself.

Consider the same for merge if the training run is successful. Resolves
[#985](https://github.com/openshiftio/openshift.io/issues/985)


rootAvish:
[test]

kbsingh:
[test]

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
```diff
@@ -23,12 +23,14 @@ def train(cls, data_store):
         return None
 
     @classmethod
-    def curate(cls, data_store, filename):
-        """Reads curated package topic dict from the given store and makes every element
-        to lower case, regenerates the package topic map and generated the topic to package map.
+    def curate(cls, data_store, filename, additional_path=""):
+        """Reads curated package topic dict from the given store and makes every
+         element to lower case, regenerates the package topic map and generated
+         the topic to package map.
```
fridex: A docstring should start with a summary line to make it consumable (here and other functions bellow) - see [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
```diff
@@ -44,12 +46,20 @@ def curate(cls, data_store, filename):
...
+                additional_path,
+                package_topic_dict))
             for package in package_topic_dict.keys():
                 topic_list = package_topic_dict[package]
                 formatted_package = package.lower()
                 formatted_topic_list = [
                     GNOSIS_PTM_TOPIC_PREFIX + x.lower() for x in topic_list]
-                distinct_formatted_topic_list = list(set(formatted_topic_list))
+                distinct_formatted_topic_list = list(
+                    set(formatted_topic_list)) \
```
fridex: You could directly use a set instead of converting set to list each time (and do casting at the end if necessary).

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
```diff
@@ -97,3 +108,41 @@ def get_dictionary(self):
...
+        manifest_file_list = data_store.list_files(prefix=additional_path +
+                                                   MANIFEST_FILEPATH)
+        unknown_packages = {}
+        for manifest_file in manifest_file_list:
+            eco_to_package_list_json_array = data_store.read_json_file(
+                manifest_file)
+            for eco_to_package_list_json in eco_to_package_list_json_array:
+                list_of_package_list = eco_to_package_list_json.get(
+                    MANIFEST_PACKAGE_LIST)
+                for package_list in list_of_package_list:
```
fridex: Using `get()` can default to None if `MANIFEST_PACKAGE_LIST` is not present. This can fail with error that None is not iterable.

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
```diff
@@ -97,3 +108,41 @@ def get_dictionary(self):
...
+        for name_part in name_parts:
+            # Exclude common parts of the package name that are not
+            # useful as tags
+            tags += [
+                GNOSIS_PTM_TOPIC_PREFIX + tag.lower()
+                for tag in name_part.split('.') if
+                tag and tag != 'com' and tag != 'org' and tag != 'io'
+            ]
+        # Make sure there are no duplicates
+        return list(set(tags[:4]))
```
fridex: Why `:4` here? Also using set and casting to list at the end can be more efficient and readable.

rootAvish: One of PGM's current limitations, each package can be associated to at most four topics. Can use a constant for this maybe, see the change. We can set it to `None` if we ever overcome this limitation.

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
rootAvish: We'll have to keep the package_topic_map assignment as a list, don't exactly recall if any part of the code explicitly requires for it to be a list so assuming the worst. Changed it so that the only part using a list is the PTM, as I gain more familiarity with the other modules in here maybe we can revisit this.

PR #96, opened 2017-10-27
Test data for unknown packages unit test

rootAvish:
There's a unit test that's currently failing as this file was missing in the code. Since they've not yet been integrated into the CI this went unnoticed.


sara-02:
@rootAvish adding the file to the above path, will fail the test `TestGnosisPackageTopicModel:test_generate_and_save_package_topic_model_local` as it picks all files from the `input-ptm-data` path.
You will have to move the folder to path `tests/data/data_gnosis/data_input_manifest_file_list/`

sara-02:
#97 will take of this.

PR #98, opened 2017-10-31
Change part of name tagging to use worpunct_tokenize

rootAvish:
We noticed that a lot of packages in maven have artifact names
of the form "something-metrics-dashboard" where the parts
of name make perfectly good tags.

We had introduced a tokenizer within the PGM itself to make sure
training does not fail due to a couple of packages missing their tags.

Combining both these approaches, I've changed the tokenizer
previously in use to tokenizer parts of the artifact and group name.


sara-02:
LGTM.

PR #100, opened 2017-10-31
Fix adding of ntlk

sara-02:
The previous change in `bootstrap.sh` is causing the following error while traning on EMR
```
2017-10-31 13:18:52,238 INFO i-03b9ec72e9ab5c2c8: new instance started
2017-10-31 13:22:54,830 ERROR i-03b9ec72e9ab5c2c8: failed to start. bootstrap action 1 failed with non-zero exit code.

```


rootAvish:
Assuming this will fix the build. There's an upstream bug with pip around dependency resolution that affected pomegranate installation as well.

PR #105, opened 2017-10-31
Introduce module wide logging

rootAvish:


sara-02:
@rootAvish Thanks.
LGTM

PR #113, opened 2017-11-09
Added openblas for faster computations

rootAvish:
The PGM code includes quite a bit of vector computations by the way of numpy arrays, this enhancement is expected to make those computations faster.


miteshvp:
LGTM. @rootAvish Just one question, will this installation alone do the trick?

rootAvish:
@miteshvp The code of the numpy/scipy libraries is written to make use of a BLAS/LAPACK implementation wherever available.

PR #115, opened 2017-11-09
Revert "Added openblas for faster computations"

rootAvish:
Reverts fabric8-analytics/fabric8-analytics-stack-analysis#113, to debug problems on stage.


sara-02:
@rootAvish thanks.

PR #116, opened 2017-11-09
Revert "Revert "Added openblas for faster computations""

rootAvish:
Reverts fabric8-analytics/fabric8-analytics-stack-analysis#115. If the staging issue turns out to be for some other reason than the addition of BLAS, merge this.


rootAvish:
This also goes in the base image.

Issue #86, opened 2017-09-28
Move all util files and function under one module.


sara-02:
Currently we have 3 `pgm_util`,`gnosis_util`, `softnet_util` in each analytics module. In addition we have a `util module` as well. 
For coherence, we should move all `util classes` and  `util functions` under the `util module`.

[gnosis util](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/gnosis/src/gnosis_util.py)
[softent util](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/softnet/src/softnet_util.py)
[pgm util](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/pgm/src/pgm_util.py)
[util module](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/tree/master/util)

PR #117, opened 2017-11-10
Refactor utils and test cases.

sara-02:


analytics_platform/kronos/src/kronos_online_scoring.py
```diff
@@ -1,6 +1,6 @@
 from analytics_platform.kronos.pgm.src.pgm_constants import *
 from analytics_platform.kronos.pgm.src.pgm_pomegranate import PGMPomegranate
-from analytics_platform.kronos.pgm.src.pgm_util import *
+from util.pgm_util import *
```
pkajaba: would you kindly change this to `import util.pgm_util as utils`. I know it is normal to use star imports in Java, but in Python, it might lead to accidentally replacing some standard methods. If you want to read more about here is interesting reading [1].

[1] https://stackoverflow.com/questions/2386714/why-is-import-bad

pkajaba: You have to also append `utils.` before every function from that namespace.

util/analytics_platform_util.py
```diff
@@ -7,3 +14,26 @@ def trunc_string_at(s, d, n1, n2):
...
+    """Create tags for a package based on its name."""
+
+    tags = []
+    name_parts = package_name.split(":")[:2]
+    if len(name_parts) > 1:
+        # ecosystem is maven, at least based on naming scheme
+        tags_artifact = [tag for tag in wordpunct_tokenize(name_parts[1]) if
+                         tag not in string.punctuation]
+        tags_group = [tag for tag in wordpunct_tokenize(name_parts[0]) if
+                      tag not in string.punctuation and tag not in
```
pkajaba: I believe you wanted to use `or` instead of `and` here :-).

util/analytics_platform_util.py
```diff
@@ -7,3 +14,26 @@ def trunc_string_at(s, d, n1, n2):
         if not result.endswith("/"):
             result += "/"
     return result
+
+
+def create_tags_for_package(package_name):
+    """Create tags for a package based on its name."""
+
+    tags = []
+    name_parts = package_name.split(":")[:2]
```
pkajaba: What if any package from other ecosystems than maven will contain some colons?

tests/data/data_softnet/output-com-data/data_co_occurrence_matrix/1/cooccurrence_matrix_maven.json
```diff
@@ -1 +1 @@
-{"p2":{"3":0,"0":1},"p3":{"3":0,"0":1},"p1":{"3":0,"0":1},"p6":{"3":1,"0":0},"p4":{"3":1,"0":0},"p5":{"3":1,"0":0},"a":{"3":1,"0":1},"b":{"3":1,"0":1},"c":{"3":1,"0":1},"d":{"3":1,"0":0},"e":{"3":1,"0":0},"f":{"3":1,"0":0},"g":{"3":0,"0":1},"h":{"3":1,"0":1},"i":{"3":0,"0":1},"j":{"3":1,"0":1},"k":{"3":1,"0":0}}
\ No newline at end of file
+{"p2":{"1":1,"3":0},"p3":{"1":1,"3":0},"p1":{"1":0,"3":0},"p6":{"1":0,"3":1},"p4":{"1":1,"3":1},"p5":{"1":0,"3":1},"a":{"1":1,"3":1},"b":{"1":1,"3":1},"c":{"1":1,"3":1},"d":{"1":1,"3":1},"e":{"1":1,"3":1},"f":{"1":1,"3":1},"g":{"1":1,"3":0},"h":{"1":1,"3":1},"i":{"1":1,"3":0},"j":{"1":1,"3":1},"k":{"1":1,"3":1}}
```
pkajaba: I can't consider whether those data are correct, so I am going to believe you :-)

pkajaba:
Can you rebase, so we don't have a merge commit?

tests/data/data_softnet/output-com-data/data_co_occurrence_matrix/1/cooccurrence_matrix_maven.json
sara-02: Actually due to random sampling, every time we run the tests this data will change. I have added it to git ignore and removed these files from appearing in future as they have no other use outside of testing.

analytics_platform/kronos/src/kronos_online_scoring.py
sara-02: Thanks I have made the necessary changes in all the modules.

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
```diff
@@ -58,12 +57,16 @@ def curate(cls, data_store, filename, additional_path=""):
...
                 topic_list = package_topic_dict[package]
+                if len(topic_list) == 0:
+                    topic_list = create_tags_for_package(package)
                 formatted_package = package.lower()
                 formatted_topic_list = [
-                    GNOSIS_PTM_TOPIC_PREFIX + x.lower() for x in topic_list][:GNOSIS_PTM_TAG_LIMIT]
-                distinct_formatted_topic_list = \
-                    set(formatted_topic_list) or cls._create_tags_for_package(package)
-                package_to_topic_dict[formatted_package] = list(distinct_formatted_topic_list)
+                    GNOSIS_PTM_TOPIC_PREFIX + x.lower() for x in
```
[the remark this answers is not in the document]
sara-02: So that we can differentiate a package_name from a tag_name/topic_name if they happend to be the same, which is required while creating a unique column for every package and topic found in the data set.

util/analytics_platform_util.py
sara-02: Will have to take ecosystem as a parameter as well.

util/analytics_platform_util.py
sara-02: Nopes, `and` is needed. Eg: if tag is `io` then using the or condition will fail.

sara-02:
@pkajaba PTAL again.

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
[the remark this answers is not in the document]
pkajaba: 1) why prefix is `c_`? 
> which is required while creating a unique column for every package and topic found in the dataset.
2) where is this uniqueness enforced? That two information are separate, so I don't see any problem if they are same. Correct me if I am wrong here.

analytics_platform/kronos/softnet/src/cooccurrence_matrix_generator.py
```diff
@@ -20,10 +20,12 @@ def __init__(self, matrix_dict):
     @classmethod
     def generate_cooccurrence_matrix(cls, kronos_dependency_dict, list_of_package_list):
 
-        kronos_intent_dependency_dict = kronos_dependency_dict.get(KD_INTENT_DEPENDENCY_MAP)
-        kronos_component_dependency_dict = kronos_dependency_dict.get(KD_COMPONENT_DEPENDENCY_MAP)
-        node_list = kronos_dependency_dict.get(KD_PACKAGE_LIST) + \
-            kronos_dependency_dict.get(KD_INTENT_LIST)
+        kronos_intent_dependency_dict = kronos_dependency_dict.get(
+            utils.KD_INTENT_DEPENDENCY_MAP)
```
pkajaba: I took a look into `utils.softnet_util` and there is no such constant defined.

sara-02: `softnet_util` is importing these constants.  I can either import them in this file as well, or reuse the import from `softnet_util`.

analytics_platform/kronos/gnosis/src/gnosis_package_topic_model.py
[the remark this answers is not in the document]
pkajaba: ok, let's say that it's required. but why is `c_`?

analytics_platform/kronos/softnet/src/cooccurrence_matrix_generator.py
pkajaba: I see but it's rather unintuitive from my point of view. Of course, it works ;-) However, you see here why we are trying to opt-out from star imports. I would suggest deleting star import from `softnet_util` and import just what you really need and here import again just what you really need from original module. 

I leave it up to you, but just let me know what you think ;-)

sara-02:
@pkajaba Thanks. 
Fixes https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/issues/86

PR #121, opened 2017-11-13
Gremlin crowdsource

sara-02:


sara-02:
For user story https://github.com/openshiftio/openshift.io/issues/1286

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
```diff
@@ -21,26 +23,35 @@ def __init__(self, pruned_package_list):
...
+        local_data_obj = LocalFileSystem(TEMP_DATA_PATH)
+
         package_list_files = input_package_topic_data_store.list_files(
             additional_path + APOLLO_INPUT_RAW_PATH)
         for package_file_name in package_list_files:
-            TagListPruner.clean_file(package_file_name,
-                                     input_package_topic_data_store,
-                                     output_package_topic_data_store,
-                                     additional_path)
+            untagged_pakcage_data = TagListPruner.clean_file(package_file_name,
```
miteshvp: typo. You may want to fix and all subsequent occurrences.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
```diff
@@ -117,25 +128,35 @@ def clean_file(cls, package_file_name,
...
+           : param package_file_name: The raw package_topic file name.
+           : param content_json_list: The raw package_topic json content.
+           : param output_package_topic_data_store: Save clean package_topic json here.
+           : param additional_path: The directory to pick the package_topic files from."""
 
         content_json_list = input_package_topic_data_store.read_json_file(
             filename=package_file_name)
+
         result_package_topic_json = []
+        untagged_pakcage_data = {}
```
miteshvp: typo again

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
```diff
@@ -117,25 +128,35 @@ def clean_file(cls, package_file_name,
...
-            pruned_package__topic_list = cls.prune_tag_list_max_count(
+            pruned_package__topic_list, untagged_packages_set = cls.prune_tag_list_max_count(
                 package_tag_list)
             temp_eco_values = {APOLLO_ECOSYSTEM: ecosystem,
                                APOLLO_PACKAGE_LIST: pruned_package__topic_list}
             result_package_topic_json.append(temp_eco_values)
-        # TODO: use singleton object, with updated package_topic_list
+            if ecosystem in untagged_pakcage_data.keys():
+                current_untagged_set = set(untagged_pakcage_data[ecosystem])
+                new_untagged_set = current_untagged_list.union(
```
miteshvp: did you intend to use `current_untagged_set` instead of the list?

miteshvp:
LGTM. Will wait for @pkajaba approval

pkajaba:
@sara-02 Please, rebase instead of merge commit.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
```diff
@@ -21,26 +23,35 @@ def __init__(self, pruned_package_list):
...
-                       additional_path):
+                       additional_path,
+                       mode="test"):
         """Generate the clean aggregated package_topic list as required by Gnosis.
 
         :param input_package_topic_data_store: The Data store to pick the package_topic files from.
         :param output_package_topic_data_store: The Data store to save the clean package_topic to.
         :param additional_path: The directory to pick the package_topic files from."""
 
+        if mode == "test":
```
pkajaba: Why do you need this `mode` in the first place? You should rename it to data_path and you don't have to have this `if`.

sara-02: Because the value of data_path is different when running the test cases, that is why mode is needed.

analytics_platform/kronos/src/config.py.template
```diff
@@ -12,6 +12,8 @@ AWS_BUCKET_NAME = os.environ.get("AWS_BUCKET_NAME","dev-stack-analysis-clean-dat
 KRONOS_SCORING_REGION = os.environ.get("KRONOS_SCORING_REGION", "")
 KRONOS_MODEL_PATH = os.environ.get("KRONOS_MODEL_PATH", KRONOS_SCORING_REGION + "/github/")
 DEPLOYMENT_PREFIX = os.environ.get("DEPLOYMENT_PREFIX", "")
-
+GREMLIN_REST_URL = "http://{host}:{port}".format(
```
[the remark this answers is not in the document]
sara-02: So that by mistake the credential  don't get committed. if someone changes their `config.py` `Config.py` is in `.gitignore`.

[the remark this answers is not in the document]
pkajaba: Oh, I can see your point now, but you are sourcing all those files from environment variables.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
pkajaba: and my point here is that you don't need to solve this through conditions. You can set this value for once This will work because you are not running tests and real code in the same instance, are you?

sara-02: So basically pass the APOLLO_PATH instead of the mode value?

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
sara-02: Not always, sometimes directly writing them in the con fig, it is easier that way, as they don't change over the testing period.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
pkajaba: pretty much, or you can have it as a class variable for example.

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
pkajaba: I would go with a way, where every developer would have a script where those secrets are stored. This script would not be in repo, might be in gitignore. This script would basically export stored secrets:

```
#!/bin/bash

export TOP_SECRET1="foo_bar"
export TOP_SECRET2="foo_bar"
.
.
.
export TOP_SECRET_N="foo_bar"
./run_actuall_code.sh
```

It's another extra script, but I find it clearer than copying configs every time.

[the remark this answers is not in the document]
sara-02: @pkajaba  best would be create an issue, as it is not limited to just this PR.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
sara-02: @pkajaba Ack, thank have updated using a temp_path variable instead of mode setting.

util/data_store/local_filesystem.py
```diff
@@ -11,10 +11,12 @@
 
 
 class LocalFileSystem(AbstractDataStore):
+
     def __init__(self, src_dir):
         self.src_dir = src_dir
         # ensure path ends with a forward slash
-        self.src_dir = self.src_dir if self.src_dir.endswith("/") else self.src_dir + "/"
+        self.src_dir = self.src_dir if self.src_dir.endswith(
```
rootAvish: You don't need this if you use `os.path.join` everywhere `src_dir` is used.

tests/unit_tests/test_apollo_tag_and_gremlin.py
```diff
@@ -0,0 +1,83 @@
...
+        output_data_store.remove_json_file(
+            filename="data_input_curated_package_topic/package_topic.json")
+
+# IMPORTANT: TestGraphUpdater needs to run after TestTagListPruner
+
+    # Test class TestGraphUpdater(TestCase):
+    def test_gremlin_updater_generate_payload(self):
+        expected_pay_load = {
+            'gremlin':
+            "g.V().has('ecosystem', 'ruby')." +
```
rootAvish: You don't need the + sign here if you use the magic string concatenation of Python.

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
pkajaba: @sara-02 Can you elaborate more?

pkajaba:
@sara-02 would you kindly rebase? :-)

pkajaba:
@sara-02 It's looking good to me, but I would appreciate some unit test for new functions.

util/data_store/local_filesystem.py
sara-02: Currently removing this will cause other tests cases to fail, this requires a module wide fix of using `os.path.join`. I will create an issue and work on PR to fix this for all files.

sara-02:
@pkajaba PTAL again.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+        """Instantiate Graph Updater."""
+        self.untagged_data = untagged_data
+
+    @classmethod
+    def generate_and_update_packages(cls, apollo_temp_path):
+        local_data_obj = LocalFileSystem(apollo_temp_path)
+        file_list = local_data_obj.list_files()
+        for file_name in file_list:
+            data = local_data_obj.read_json_file(file_name)
+            # TODO: use a singleton object with updated datafile.
```
pkajaba: Can you implement this as a singleton? Anyway, I don't really like that you are initializing the instance of the object inside of the same class as an initalized object is.

Is it some design pattern or what is the reason behind it?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+            graph_obj.update_graph()
+            local_data_obj.remove_json_file(file_name)
+
+    def update_graph(self):
+        """For each ecosystem, update the graph with packages that need tags."""
+
+        for each_ecosystem in self.untagged_data:
+            package_list = self.untagged_data[each_ecosystem]
+            pck_len = len(package_list)
+            if pck_len == 0:
```
pkajaba: I would delete this `if` because it not required in context for this method.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+                # If pck_len =0 then, no package of that ecosystem requires
+                # tags. Hence, do nothing.
+                continue
+            for index in range(0, pck_len, 100):
```
pkajaba: is `100` really value what you want to have here? it means that just every 100th element in range will be used.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+                sub_package_list = package_list[index:index + 100]
+                pay_load = self.generate_payload(
```
pkajaba: why do you have `underscore` here?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+                    each_ecosystem, sub_package_list)
+                self.execute_gremlin_dsl(pay_load)
+
+    def generate_payload(self, ecosystem, package_list):
```
pkajaba: this method can be class/static.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+        """Generate the payload that is sent to graph.
+
+            :param ecosystem: The ecosystem whose packages are to be updated.
+            :param package_list: All the packages ina givne ecosystem that require tags.
+
+            :return pay_load: The formated gremlin query."""
+
+        # NOTE: This query assumes that the packages in the package_list are already
```
pkajaba: what does this mean? what if packages are not in the graph? I don't see any error handling or retry.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+            ".has('name', within(str_packages))." + \
+            "property('manual_tagging_required', true).valueMap();"
+        pay_load = {
+            'gremlin': str_gremlin_query,
+            'bindings': {'str_packages': package_list}
+        }
+        return pay_load
+
+    def execute_gremlin_dsl(self, pay_load):
+        """Execute the gremlin query submitted in the pay load"""
```
pkajaba: We have a check_style script inside. Could you run it at least for new files and fix warnings?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+        str_gremlin_query = "g.V().has('ecosystem', '{}')".format(ecosystem) + \
+            ".has('name', within(str_packages))." + \
+            "property('manual_tagging_required', true).valueMap();"
+        pay_load = {
+            'gremlin': str_gremlin_query,
+            'bindings': {'str_packages': package_list}
+        }
+        return pay_load
+
+    def execute_gremlin_dsl(self, pay_load):
```
pkajaba: this can be class/static method too.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+        """Execute the gremlin query submitted in the pay load"""
+        try:
+            response = get_session_retry().post(
+                GREMLIN_REST_URL, data=json.dumps(pay_load))
+            if response.status_code != 200:
+                _logger.error(
+                    "The rest failed with payload::\n {}".format(pay_load))
```
pkajaba: it might be good to provide more information about what error happened.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+
+        except Exception as e:
```
pkajaba: I would be great if you weakened this Exception.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
```diff
@@ -21,26 +21,33 @@ def __init__(self, pruned_package_list):
...
-            TagListPruner.clean_file(package_file_name,
-                                     input_package_topic_data_store,
-                                     output_package_topic_data_store,
-                                     additional_path)
+            untagged_package_data = TagListPruner.clean_file(package_file_name,
+                                                             input_package_topic_data_store,
+                                                             output_package_topic_data_store,
+                                                             additional_path)
+            local_data_obj.write_json_file(
+                package_file_name.split("/")[-1], untagged_package_data)
```
pkajaba: what does this splitting do? is it really required? If so, can you elaborate why?

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
pkajaba: @sara-02 ^^

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+
+class GraphUpdater(object):
+    """Update the Crowsouring graph with packages that need tags."""
+
+    def __init__(self, untagged_data):
+        """Instantiate Graph Updater."""
+        self.untagged_data = untagged_data
+
+    @classmethod
+    def generate_and_update_packages(cls, apollo_temp_path):
```
pkajaba: You have some tests but you are not testing this method. Any specific reason?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
```diff
@@ -0,0 +1,75 @@
...
+        local_data_obj = LocalFileSystem(apollo_temp_path)
+        file_list = local_data_obj.list_files()
+        for file_name in file_list:
+            data = local_data_obj.read_json_file(file_name)
+            # TODO: use a singleton object with updated datafile.
+            graph_obj = GraphUpdater(untagged_data=data)
+            graph_obj.update_graph()
+            local_data_obj.remove_json_file(file_name)
+
+    def update_graph(self):
```
pkajaba: The same comment about testing applies here.

tests/unit_tests/test_apollo_tag_and_gremlin.py
```diff
@@ -0,0 +1,82 @@
...
+        gremlin_query += "has('name', within(str_packages))."
+        gremlin_query += "property('manual_tagging_required', true).valueMap();"
+        expected_pay_load = {
+            'gremlin': gremlin_query,
+
+            'bindings': {
+                'str_packages': ['service_identity']}}
+
+        unknown_data_obj = LocalFileSystem(APOLLO_TEMP_TEST_DATA)
+        self.assertTrue(unknown_data_obj is not None)
```
pkajaba: `assert unknown_data_obj` would be enough here.

tests/unit_tests/test_apollo_tag_and_gremlin.py
```diff
@@ -0,0 +1,82 @@
...
+        for tag_generated, tag_expected in zip(output_result, expected_output_result):
+            self.assertDictEqual(tag_generated, tag_expected)
+
+        output_data_store.remove_json_file(
+            filename="data_input_curated_package_topic/package_topic.json")
+
+# IMPORTANT: TestGraphUpdater needs to run after TestTagListPruner
+
+    # Test class TestGraphUpdater(TestCase):
+    def test_gremlin_updater_generate_payload(self):
```
pkajaba: what does this method is supposed to test? you don't really have read package list from the file system, just create a fixture of package list and tests whether a query is created correctly.

tests/unit_tests/test_apollo_tag_and_gremlin.py
```diff
@@ -0,0 +1,82 @@
...
+from analytics_platform.kronos.apollo.src.apollo_constants import APOLLO_TEMP_TEST_DATA
+from analytics_platform.kronos.apollo.src.apollo_gremlin import GraphUpdater
+
+from unittest import TestCase
+
+
+class TestPruneAndUpdate(TestCase):
+
+    # Test Class TagListPruner
+    def test_generate_and_save_pruned_list_local(self):
```
pkajaba: I am struggling to see what method is this test testing. Can you elaborate?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: @pkajaba This method talks to gremlin, so every time we are testing it we need to a working instance of grmelin-http up and running.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Same reason.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Yes, this can be passed as parameter, but that is what is intended, as we want to break the list in chunks of 100 subsets each.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: I don't see any dangling underscore :/

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack

tests/unit_tests/test_apollo_tag_and_gremlin.py
sara-02: The input list contains more than 4 tags, the prune method will generate tag list upto 4 tags based on frequency. So this test checks that the desired 4 tags are generated or not.

tests/unit_tests/test_apollo_tag_and_gremlin.py
sara-02: Ack, I don't need to load the list, but generation of list needs to be checked, so I am adding it to the previous test instead of this one.

tests/unit_tests/test_apollo_tag_and_gremlin.py
sara-02: Ack

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: This query will execute even if the package is not in the graph, no excepction will be thrown by the graph. But no update on that package will take place.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack

analytics_platform/kronos/apollo/src/apollo_gremlin.py
pkajaba: a payload is the correct word, right?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
pkajaba: ack

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack.

tests/unit_tests/test_apollo_tag_and_gremlin.py
pkajaba: yeah, you can spit it, but if a generation of the list really has to be tested it should be extracted in function.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: I think we can skip this try-catch as it should be taken care by the session itself, [here](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/pull/121/files#diff-1731aab880df79a3abc148fddb240110)

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
sara-02: How will it get the configs inside the Docker file?
https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/Dockerfile#L24

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
sara-02: The package_name is the absolute name something like `maven/github/data_input_raw_package_list/package_topic.json` When I save the pruned `package_topic.json` it will be `new_location/package_topic.json` That is why split is needed.

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
pkajaba: through environment variables.

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
pkajaba: is this name from S3? from your example there no clear name of the package. Was it just bad example or package_name is not really package name?

tests/unit_tests/test_apollo_tag_and_gremlin.py
pkajaba: but which method does it test in code? I can't find `test_generate_and_save_pruned_list_local` method in repository. 

Unit tests should test methods and their behavior on various inputs.

analytics_platform/kronos/src/config.py.template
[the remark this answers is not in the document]
sara-02: Ok will add a PR for that separately.

tests/unit_tests/test_apollo_tag_and_gremlin.py
pkajaba: @sara-02

tests/unit_tests/test_apollo_tag_and_gremlin.py
pkajaba: @sara-02 ^^

analytics_platform/kronos/apollo/src/apollo_gremlin.py
pkajaba: so I would advise here to mock HTTP response to emulate the behavior of graph DB.

analytics_platform/kronos/apollo/src/apollo_gremlin.py
sara-02: Ack

analytics_platform/kronos/apollo/src/apollo_tag_prune.py
sara-02: No from the local file, the package_name contains the absolute name, because we have load its contents.

tests/unit_tests/test_apollo_tag_and_gremlin.py
pkajaba: @sara-02 ^^ :-)

analytics_platform/kronos/apollo/src/apollo_gremlin.py
pkajaba: can you elaborate more on how it should be taken care by session?

analytics_platform/kronos/apollo/src/apollo_gremlin.py
pkajaba: Well, we should write into logs that something like happened, right?

tests/unit_tests/test_apollo_tag_and_gremlin.py
sara-02: I am testing extraction here itself and then deleting the files. If not then it will create a dependency of tests as we have make sure that the extraction always gets tested after generation if they are 2 separate tests.

sara-02: This function only checks for payload now.

util/data_store/local_filesystem.py
rootAvish: I already fixed it for all files that were present in the source code when I did my Python3 changes.

rootAvish: The "apollo" module should be the only problem, ergo not too hard to fix.

PR #123, opened 2017-11-19
Log response time

msrb:


msrb:
Note I am not sure whether `teardown_request` is the right place where to stop counting. Let's test this and see...

msrb:
@sara-02 PTAL, thanks 😉

sara-02:
Hey @msrb when I ran it locally with docker-compose then I am not able to see the time being logged, but if I set the log level to error 
```
app.logger.error('Response time: {t} seconds'.format(
        t=(datetime.datetime.utcnow() - g.request_start).total_seconds()))
```
then I see something like this `kronos_1  | [2017-11-20 13:56:52,168] ERROR in rest_api: Response time: 16.758821 seconds` :/

msrb:
@sara-02 please take a look now.

This should do the trick:

```bash
sudo docker run -e SERVICE_PORT=5000 -e SERVICE_TIMEOUT=30 -e FLASK_LOGGING_LEVEL=DEBUG  -t --rm -p 5000:5000 registry.devshift.net/bayesian/kronos:SNAPSHOT-PR-123
```

For staging and production, we can set the logging logging level in https://github.com/openshiftio/saas-analytics/blob/master/bay-services/kronos.yaml#L9-L11.

sara-02:
@msrb  Thanks.
LGTM

PR #135, opened 2017-11-30
[WIP] Dockerfile changes to migrate to Python3

rootAvish:
Only merge if #45 is in and the latest version of pomegranate is verified to work on Python3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,20 +3,23 @@ MAINTAINER Avishkar Gupta <avgupta@redhat.com>
 
 RUN yum install -y epel-release && \
     yum install -y zip && \
-    yum install -y python-pip python-devel gcc && \
+    yum install -y gcc && \
     yum -y install tkinter && \
     yum-config-manager --disable testing-devtools-2-centos-7 && \
     yum -y install gcc-c++.x86_64 && \
     yum clean all
 
+RUN yum -y install https://centos7.iuscommunity.org/ius-release.rpm
+RUN yum -y install python36u && \
+    yum -y install python36u-pip && \
+    yum -y install python36u-devel
 
 # --------------------------------------------------------------------------------------------------
 # install python packages
 # --------------------------------------------------------------------------------------------------
 COPY ./analytics_platform/kronos/requirements.txt /
-RUN pip install -r /requirements.txt && rm /requirements.txt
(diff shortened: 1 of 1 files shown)


rootAvish:
Only merge if #45 is in and the latest version of pomegranate is verified to work on Python3.

Dockerfile
```diff
@@ -3,20 +3,23 @@ MAINTAINER Avishkar Gupta <avgupta@redhat.com>
...
 RUN yum install -y epel-release && \
     yum install -y zip && \
-    yum install -y python-pip python-devel gcc && \
+    yum install -y gcc && \
     yum -y install tkinter && \
     yum-config-manager --disable testing-devtools-2-centos-7 && \
     yum -y install gcc-c++.x86_64 && \
     yum clean all
 
+RUN yum -y install https://centos7.iuscommunity.org/ius-release.rpm
```
pkajaba: install epel-release

Dockerfile
```diff
@@ -3,20 +3,23 @@ MAINTAINER Avishkar Gupta <avgupta@redhat.com>
...
+RUN yum -y install python36u && \
```
pkajaba: and from epel-release install python34 please. On top of that you can install it all in one yum command instead of 3.

rootAvish: I've not tested it on 3.4, are there any major differences between the two releases?

Dockerfile
rootAvish: Again, this is there because I wanted 3.6. I'm not sure about 3.4 but I can test.

pkajaba: Firstly, do you have any experience with `centos7.iuscommunity.org`? I am seeing it for the first time. If you do, I am sorry to ask. 

Secondly, we are using epel release of Python 3.4 in core stream, so it would be nice to be consistent.

Dockerfile
pkajaba: nope, please read the previous comment. There are no major changes, which would impact our code.

rootAvish: Cool, I'll still like to test once since Kronos doesn't have good enough test coverage for me to just run the test suite and be sure that nothing is breaking. If it works I'll put in all the changes you suggested.

rootAvish:
Closing this as we now have a base image and this change needs to go there.

PR #137, opened 2017-12-01
Unit test integration in the CI pipeline

rootAvish:
Okay, so Spark installation doesn't take as long as I thought it would, however the bottleneck in building all of this is the Kronos image build itself since it has a lot of Cpython/C/Fortran code in the numpy/scipy suite and Pomegranate that needs to build wheel on the container.

Is it possible for us to push to a 'pomegranate' base image to registry.devshift, that contains all the Python dependencies for Kronos in it? That way we reduce the time it takes to build Kronos and in turn the time taken by the CI pipeline.

### DIFF
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -13,7 +13,7 @@ RUN yum install -y java-1.8.0-openjdk.x86_64 java-1.8.0-openjdk-devel.x86_64 wge
     wget https://downloads.lightbend.com/scala/2.12.2/scala-2.12.2.tgz && \
     tar -xvf scala-2.12.2.tgz && \
     mv scala-2.12.2 /usr/local/scala && \
-    wget https://d3kbcqa49mib13.cloudfront.net/spark-2.1.1-bin-hadoop2.7.tgz && \
+    wget https://archive.apache.org/dist/spark/spark-2.1.1/spark-2.1.1-bin-hadoop2.7.tgz && \
     tar -xvf spark-2.1.1-bin-hadoop2.7.tgz && \
     mv spark-2.1.1-bin-hadoop2.7 /usr/local/spark
 
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -1,0 +1,31 @@
+FROM registry.devshift.net/bayesian/kronos:latest
+MAINTAINER Avishkar Gupta <avgupta@redhat.com>
+
+# --------------------------------------------------------------------------------------------------
+# copy testing source code and scripts into root dir /
+# --------------------------------------------------------------------------------------------------
+ADD ./tests/ /tests
+ADD ./tests/scripts/entrypoint-test.sh /entrypoint-test.sh
+RUN chmod +x /entrypoint-test.sh
+
+# Begin spark installation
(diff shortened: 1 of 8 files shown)


Dockerfile.tests
```diff
@@ -0,0 +1,31 @@
...
+ADD ./tests/ /tests
+ADD ./tests/scripts/entrypoint-test.sh /entrypoint-test.sh
+RUN chmod +x /entrypoint-test.sh
+
+# Begin spark installation
+RUN yum install -y java-1.8.0-openjdk.x86_64 java-1.8.0-openjdk-devel.x86_64 wget && \
+    wget https://downloads.lightbend.com/scala/2.12.2/scala-2.12.2.tgz && \
+    tar -xvf scala-2.12.2.tgz && \
+    mv scala-2.12.2 /usr/local/scala && \
+    wget https://d3kbcqa49mib13.cloudfront.net/spark-2.1.1-bin-hadoop2.7.tgz && \
```
pkajaba: any specific reason not to use official download for spark? [1]

[1] https://spark.apache.org/downloads.html

pkajaba:
@rootAvish It's possible. We are basically doing the same thing with worker image. Feel free to take a look here: https://github.com/fabric8-analytics/fabric8-analytics-worker-base

runtest.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+TEST_IMAGE_NAME='kronos-tests'
+docker run ${TEST_IMAGE_NAME}
```
pkajaba: @rootAvish this is OK in terms of CI, since we are sure that it will go through the whole pipeline and images will be built. However, locally it's not working. Do you think that we could tinker with this script so it builds test image here?

It's just suggestion, though :-)

Dockerfile.tests
rootAvish: It's the mirror that the "official" Jupyter spark notebook container uses: https://github.com/jupyter/docker-stacks/blob/master/pyspark-notebook/Dockerfile#L18, I assumed it had some prebuilt components, I'll try linking against the official release to see if it works.

rootAvish:
@pkajaba I have a doubt there, does creating a repository under fabric8-analytics automatically add it to centos-ci, or do I need somebody with access to add that repository?

msrb:
@rootAvish please take a look here: https://github.com/fabric8-analytics/fabric8-analytics-common/blob/master/README-cicd.md#configure-cicd-jobs-in-almightyalmighty-jobs

Dockerfile.tests
pkajaba: OK, thanks for an explanation. I was just curious.

tests/scripts/entrypoint-test.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+python -m pytest /tests/unit_tests/
```
sara-02: @rootAvish locally the command that is working is `python -m pytest tests/unit_tests/`

rootAvish: That's because locally you don't copy the tests to / 🙂 . Locally you can just proceed by `./runtest.sh` once it's done.

rootAvish:
@pkajaba Thanks for the review, guess things look better now.

PR #138, opened 2017-12-05
Fix for broken unit tests

rootAvish:
All the more reason to have these in the pipeline 🙂 , luckily these did not break anything in production as the model there does not use the local filesystem.


sara-02:
LGTM

PR #140, opened 2017-12-07
Remove gevent as worker class to remove nested multiprocessing

abs51295:
Gevent spans a new process for every worker and since PGM also performs scoring in [parallel](https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/pgm/src/pgm_pomegranate.py#L82-L86), it results into nested multiprocessing which isn't supported by joblib. According to this joblib [issue](https://github.com/joblib/joblib/issues/180), it converts nested calls into sequential code which defeats the whole purpose of multiprocessing and doesn't allow PGM to perform the scoring in parallel which in turn makes PGM much slower. 

By removing gevent, PGM is now able to span multiple processes and hence the scoring is approximately 4-5 times faster than the original one. :tada:

### DIFF
--- a/analytics_platform/kronos/requirements.txt
+++ b/analytics_platform/kronos/requirements.txt
@@ -14,7 +14,6 @@ boto3==1.4.4
 scipy==0.19.0
 botocore==1.5.32
 pandas==0.19.2
-gevent==1.2.2
 nltk==3.2.5
 daiquiri==1.3.0
 six==1.11.0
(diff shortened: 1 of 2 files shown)


msrb:
@abs51295 @rootAvish 

A bit late to the party, I know :) But I am curious here. Does this mean that gunicorn is now configured in "sync workers" mode and it can process only 2 requests in parallel?

abs51295:
Yes @msrb you are correct.

rootAvish:
We can always revert to using gevent if joblib ever starts to support nested multi-processing.

PR #149, opened 2017-12-11
Fixed gnosis package topic model generation overwrite bug

rootAvish:
If we have two packages of the form:

"mysql:mysql-connector-java": ['tag1', 'tag2']
"MySql:mysql-connector-java": []

Then the auto generated tags from the part of name tagger overwrite the first. as they're both mapped to the same lowercase key.


sara-02:
@rootAvish I am merging this PR, will add unit test separately, in the PR that I am working on.

PR #154, opened 2017-12-14
Coccurrence Count + confidence added.

sara-02:
* Replace the value of simply co-occurrence count with a confidence count ratio. 
* Add another key-vlaue to hold the numerical co-occurrence count.

### DIFF
--- a/analytics_platform/kronos/src/recommendation_validator.py
+++ b/analytics_platform/kronos/src/recommendation_validator.py
@@ -160,12 +160,15 @@ class RecommendationValidator(object):
 
         final_companion_recommendations = []
         count = Counter()
+        input_freq_count = float(self.check_companion_or_alternate_recommendation_validity(
+            set(input_list)))
         for each_recommendation in companion_packages:
             companion_package = each_recommendation.get('package_name')
             recommended_dependency_set = self.generate_companion_dependency_set(
                 input_list, companion_package)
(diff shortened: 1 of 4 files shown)


sara-02:
@miteshvp yes then both score and occurrence count will get removed, and will require subsequent change in the UI work flow as well.

analytics_platform/kronos/src/recommendation_validator.py
```diff
@@ -160,22 +160,29 @@ def check_companion_recommendation(self, input_list, companion_packages, top_cou
...
         for each_recommendation in companion_packages:
             companion_package_name = each_recommendation.get('package_name')
             comp_count = count[companion_package_name]
             if companion_package_name in top_comp_name and comp_count:
-                # change the value of cooccurrence_probability to reflect the actual manifest
-                # usage count.
-                each_recommendation["cooccurrence_probability"] = comp_count
+                """change the value of cooccurrence_probability to reflect
+                the actual manifest confidence count.
+                Confidence count = NumOfOccurence(input+companion)/NumOfOccurence(input)"""
```
pkajaba: @sara-02 what does this line do? It looks like some Java code to me.

sara-02: Just a comment to show how the confidence count is calculated. As there can be many ways of implementing it.

PR #156, opened 2017-12-20
Proper folder path for os.path.join

sara-02:


util/data_store/local_filesystem.py
```diff
@@ -11,10 +11,9 @@
 
 
 class LocalFileSystem(AbstractDataStore):
+
     def __init__(self, src_dir):
         self.src_dir = src_dir
-        # ensure path ends with a forward slash
-        self.src_dir = self.src_dir if self.src_dir.endswith("/") else self.src_dir + "/"
```
rootAvish: Other than removing this line, you actually didn't need to make any other changes. But since you worked so hard on these I won't ask you to revert them.

rootAvish:
@sara-02 You should look into the documentation of this function once to see why I asked you to use it, you actually didn't need to make these changes. 🙂

PR #161, opened 2017-12-29
Fix for duplicate input packages

### LINKED ISSUE #160: Empty Response when user input has duplicate packages
When a request is sent to PGM with duplicate packages, the response received is empty. For example:

Request: 
```json
[
    {
        "ecosystem": "maven",
        "comp_package_count_threshold": 5,
        "alt_package_count_threshold": 2,
        "outlier_probability_threshold": 0.88,
        "unknown_packages_ratio_threshold": 0.3,
        "package_list": [
          "io.vertx:vertx-core",
          "io.vertx:vertx-core",
          "io.vertx:vertx-web"
       ]
    }
]		
```

Response:
```json
[
    {
        "alternate_packages": {},
        "companion_packages": [],
        "ecosystem": "maven",
        "missing_packages": [],
        "outlier_package_list": [],
        "package_to_topic_dict": {},
        "user_persona": "1"
    }
]
```

abs51295:
Resolves #160 .

### DIFF
--- a/analytics_platform/kronos/src/recommendation_validator.py
+++ b/analytics_platform/kronos/src/recommendation_validator.py
@@ -203,20 +203,18 @@ class RecommendationValidator(object):
                     only_outlier_alternate[package] = alternate_package[package]
         return only_outlier_alternate
 
-    def get_filtered_input_list(self, input_list, missing_packages):
+    def get_filtered_input_list(self, input_set, missing_packages):
         """Prune the original package list to exclude the missing packages for
         testing recommendation validation.
 
-        :param input_list: The original package list of the user stack.
+        :param input_set: The original package set of the user stack.
         :param missing_packages: The list of Packages unknown to PGM.
 
         :return : Filtered input package list."""
 
         filtered_input_list = []
-        if input_list is not None and missing_packages is not None:
-            missing_packages_set = set(missing_packages)
-            filtered_input_list = [
-                package for package in input_list if package not in missing_packages_set]
+        if input_set is not None and missing_packages is not None:
+            filtered_input_list = list(input_set.difference(missing_packages))
         return filtered_input_list
 
     def get_all_list_package_length(self):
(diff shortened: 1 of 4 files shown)


rootAvish:
Just one critique here, can you add a unit test for this so I know it's working?

abs51295:
Sure @rootAvish :+1:

rootAvish:
@abs51295 It does not 🙂 , I guess you should convert the set back to a list after you've filtered duplicates.

abs51295:
Yes, need to change the test case to include set and not list.

rootAvish:
Have the changes been made in consumers elsewhere or were they not required?

rootAvish:
Inside the codebase, not the tests.

abs51295:
Changes have been made in the consumers elsewhere. Only need to change the tests to check the same.

analytics_platform/kronos/src/recommendation_validator.py
```diff
@@ -203,20 +203,18 @@ def get_filtered_alternate_list(self, alternate_package, outlier_packages):
...
 
         :return : Filtered input package list."""
 
         filtered_input_list = []
-        if input_list is not None and missing_packages is not None:
-            missing_packages_set = set(missing_packages)
-            filtered_input_list = [
-                package for package in input_list if package not in missing_packages_set]
+        if input_set is not None and missing_packages is not None:
+            filtered_input_list = list(input_set.difference(missing_packages))
```
sara-02: Are you sure about this? `missing_packages` is a list not set.

sara-02: Looks like it gets taken care by the `difference` function internally.

abs51295: Yes @sara-02 you can check the [docs](https://docs.python.org/2/library/sets.html) for the clarification. It clearly says this `Note, the non-operator versions of update(), intersection_update(), difference_update(), and symmetric_difference_update() will accept any iterable as an argument.`

sara-02: Yes, I tried with the `-` for difference operation, and it was throwing an error. But `difference` function is working fine :)

abs51295: Thanks :)

abs51295:
@rootAvish I have added the test and they have successfully passed :)

PR #163, opened 2018-01-02
Remove companion thresholding before filter and retrain

rootAvish:
There's already a thresholding step implemented in the post-processing
filters where we only choose the top three companion recommendations.
The current model is using 2-itemsets and works for all the quickstarts
when coupled with the changes in data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,4 +159,4 @@ BODY: JSON data
 ## Latest Depolyment
 
 * Maven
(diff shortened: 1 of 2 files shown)


rootAvish:
Solves https://github.com/openshiftio/openshift.io/issues/1500 and https://github.com/openshiftio/openshift.io/issues/1450.

PR #164, opened 2018-01-03
Remove pickling of model

rootAvish:
The model is already being converted to a JSON(using the `to_json`)
function of pomegranate which is already a stadard serialized format.
I don't see a need to further serialize the JSON using pickle to
something that can be loaded only using Python. This also helps us
reduce the training time of the model as pickling and unpickling
has a overhead that I don't see a need for, because JSON.


miteshvp:
LGTM. Will wait for green from Jenkins

PR #167, opened 2018-01-03
Use replica varibale instead of constant

sara-02:


openshift/template.yaml
```diff
@@ -14,7 +14,7 @@ objects:
       service: bayesian-kronos-${KRONOS_SCORING_REGION}
     name: bayesian-kronos-${KRONOS_SCORING_REGION}
   spec:
-    replicas: 1
+    replicas: ${{REPLICAS}}
```
miteshvp: I guess you either need ${} or {{}} but not both

sara-02: @mitesh not sure myself, I followed the expression here: https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/openshift/template.yaml#L154

miteshvp: can you please make it `"${{REPLICAS}}"`

sara-02: If we put it under " " will it not be considered a string, or will it still do the replacement?

sara-02:
After reading some stuff as to why `{{}}` parameter was used, instead of the usual `{}` one, the only logical conclusion is that we want the replica parameter to be a numeric value, so we use `{{}}` as a hack to get unquoted data.
As per the documentation 
`
When using the ${{PARAMETER_NAME}} syntax only a single parameter reference is allowed and leading/trailing characters are not permitted. The resulting value will be unquoted unless, after substitution is performed, the result is not a valid json object. If the result is not a valid json value, the resulting value will be quoted and treated as a standard string.
`
Source : https://docs.openshift.org/latest/dev_guide/templates.html
@miteshvp if we used `""` then it might end up as string parameter.  Do we want that?
@msrb is this the reason for using `{{}}` for replica parameter?

miteshvp:
@sara02 - In the same source you provided, please look at Example-7 Bullet Point 2. Yes, with the latest Openshift, it requires all the Params to be string literals.

sara-02:
@miteshvp ack :), will make the updates.

PR #168, opened 2018-01-04
Take outlier logic outside of PGM

abs51295:
The current outlier logic wasn't scaling and hence a new logic was proposed and set into place.


analytics_platform/kronos/apollo/src/offline_training.py
```diff
@@ -1,5 +1,6 @@
 from analytics_platform.kronos.apollo.src.apollo_tag_prune import TagListPruner
 from analytics_platform.kronos.src import config
+from apollo_generate_frequency_dict import FrequencyDictGenerator
```
sara-02: Either use full path, or prefix a `.`, to keep the path compatible with both Python2 and 3

analytics_platform/kronos/deployment/rest_api.py
```diff
@@ -14,7 +14,8 @@
 from analytics_platform.kronos.src.config import (
     AWS_BUCKET_NAME, KRONOS_MODEL_PATH, KRONOS_SCORING_REGION)
 from analytics_platform.kronos.src.kronos_online_scoring import \
-    load_user_eco_to_kronos_model_dict_s3, score_eco_user_package_dict
+    load_user_eco_to_kronos_model_dict_s3, score_eco_user_package_dict, \
```
sara-02: Not specific to this PR, but it would be more consistent if we used `()` instead of `\` for multi-imports.

tests/unit_tests/test_kronos_pgm_pomegranate.py
```diff
@@ -1,10 +1,12 @@
-from analytics_platform.kronos.src import config
 from analytics_platform.kronos.pgm.src.pgm_pomegranate import PGMPomegranate
 from analytics_platform.kronos.pgm.src.offline_training import load_eco_to_kronos_dependency_dict, \
     load_user_eco_to_co_occerrence_matrix_dict
 from analytics_platform.kronos.src.kronos_online_scoring import score_eco_user_package_dict, \
     load_user_eco_to_kronos_model_dict, get_eco_to_kronos_dependency_dict
+from analytics_platform.kronos.pgm.src.pgm_constants import KD_PACKAGE_FREQUENCY
+from analytics_platform.kronos.src.recommendation_validator import RecommendationValidator
 from util.data_store.local_filesystem import LocalFileSystem
+from pprint import pprint
```
sara-02: Do we need this import and its subsequent usage?

sara-02:
Good work @abs51295

util/analytics_platform_util.py
```diff
@@ -34,3 +36,18 @@ def get_path_names(training_data_url):
     output_bucket_name = trunc_string_at(training_data_url, "/", 2, 3)
     additional_path = trunc_string_at(training_data_url, "/", 3, -1)
     return input_bucket_name, output_bucket_name, additional_path
+
+
+def load_package_list(input_data_store, additional_path):
```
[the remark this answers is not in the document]
abs51295: Done @sara-02 :)

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
```diff
@@ -0,0 +1,45 @@
...
+    """Generates package to frequency dictionary"""
+
+    def __init__(self, manifest_file):
+        """Initialize FrequencyDictGenerator with manifest_file of ecosystem.
+
+           :param manifest_file: The manifest.json file for intended ecosystem. """
+        self.manifest_file = manifest_file
+
+    @classmethod
+    def read_manifest_file(cls, input_data_store, additional_path):
```
rootAvish: If this is a factory method, I would like for it to named as such and the docstring to contain that this code generates an instance of this class.

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
```diff
@@ -0,0 +1,45 @@
...
+
+        :param output_data_store: Data store to write the frequency dict to.
+        :param additional_path: Path to store the frequency dict inside the output_data_store.
+        """
+        package_to_frequency_dict = dict()
+
+        for package_set in self.manifest_file:
+            for package in package_set:
+                if package_to_frequency_dict.get(package) is not None:
+                    package_to_frequency_dict[package] += 1
```
rootAvish: You can create this using `collections.Counter`. I would prefer you do that, it gets rid of this `if-else` condition.

rootAvish: On that note, why is a custom class required here at all?

analytics_platform/kronos/deployment/rest_api.py
```diff
@@ -122,7 +126,8 @@ def predict_and_score():
             user_request=input_json,
             user_eco_kronos_dict=app.user_eco_kronos_dict,
             eco_to_kronos_dependency_dict=app.eco_to_kronos_dependency_dict,
-            all_package_list_obj=app.all_package_list_obj)
+            all_package_list_obj=app.all_package_list_obj,
+            package_frequency_dict=app.package_frequency_dict)
```
rootAvish: Why do you need to store this in the app context?

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
```diff
@@ -83,7 +83,7 @@ def score(self, evidence_dict_list):
         n_jobs = len(evidence_dict_list)
         result_array = functools.reduce(
             list.__add__,
-            Parallel(n_jobs=n_jobs)(
+            Parallel(n_jobs=n_jobs, verbose=51)(
```
rootAvish: I don't think this is required in production, is this a leftover from your performance testing branch?

analytics_platform/kronos/src/kronos_online_scoring.py
```diff
@@ -60,6 +75,15 @@ def get_sorted_companion_intent_probabilities(res, node_list):
     return result
 
 
+def get_topics_for_manifest(manifest, package_topic_dict):
+    manifest_topics = set()
```
rootAvish: This entire function can be compressed to a one line set comprehension.

analytics_platform/kronos/src/kronos_online_scoring.py
```diff
@@ -129,35 +192,34 @@ def get_clean_topics_for_package(package_to_topic_dict, package):
...
-        outlier_result = result_array[i][potential_outlier_index_list[i]]
-        outlier_score = outlier_result.values()[0]
-        if outlier_score > outlier_threshold:
-            outlier_dict = dict()
-            outlier_dict[pgm_constants.KRONOS_OUTLIER_PACKAGE_NAME] = observed_package_list[i]
-            outlier_dict[pgm_constants.KRONOS_OUTLIER_PROBABILITY] = outlier_score
-            outlier_dict[pgm_constants.KD_TOPIC_LIST] = \
-                get_clean_topics_for_package(package_to_topic_dict, observed_package_list[i])
-            outlier_dict_list.append(outlier_dict)
+        outlier_dict = dict()
```
rootAvish: I don't like this format over the normal initialization.

```python
outlier_dict = {
        pgm_constants.KRONOS_OUTLIER_PACKAGE_NAME: potential_outlier_index_list[i],
        pgm_constants.KRONOS_FREQUENCY_COUNT : int(
            package_frequency_dict.get(potential_outlier_index_list[i], 0)),
        pgm_constants.KD_TOPIC_LIST:
            get_clean_topics_for_package(package_to_topic_dict, potential_outlier_index_list[i])
}
```

tests/unit_tests/test_kronos_pgm_pomegranate.py
```diff
@@ -94,53 +109,64 @@ def test_score_eco_user_package_dict(self):
...
+            input_folder_name="tests/data/data_recom_valid/",
+            additional_path="")
+
+        response = score_eco_user_package_dict(
+            user_request, user_eco_kronos_dict=user_eco_kronos_dict,
+            eco_to_kronos_dependency_dict=eco_to_kronos_dependency_dict,
+            all_package_list_obj=all_package_list_obj,
+            package_frequency_dict=frequency_dict)
+
+        self.assertTrue(response is not None)
```
rootAvish: There's a method `assertIsNotNone`.

util/analytics_platform_util.py
```diff
@@ -34,3 +36,24 @@ def get_path_names(training_data_url):
...
+    """Loads the manifest files from the input_data_store and
+    returns an aggregated list of set of packages.
+
+    :param input_data_store: Data store to read the manifest files from.
+    :param additional_path: Indicates path inside the data store for the manifests.
+    :return: list of package set."""
+    all_list_of_package_set = list()
+    manifest_filenames = input_data_store.list_files(
```
rootAvish: `os.path.join` here.

analytics_platform/kronos/src/kronos_online_scoring.py
rootAvish: If you must have a function, still make it a one liner.

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
abs51295: I thought it might be good to have a custom class to make it extensible in the future.

abs51295: `collections.Counter` approach would look something like `Counter(x for sublist in my_list for x in sublist)` which doesn't look readable to me and hence I followed this approach.

analytics_platform/kronos/deployment/rest_api.py
abs51295: Because it has to be loaded only once and needs to be sent with each request.

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
abs51295: I thought this might be helpful to see the time taken by PGM in staging and production, so that we don't rely on assumptions that what the actual bottleneck is.

analytics_platform/kronos/src/kronos_online_scoring.py
abs51295: One liner set comprehension is a good idea, but I am more inclined towards the readability and understandability of the code. Making it a one liner would result in a nested generator expression which I think would not help much.

analytics_platform/kronos/src/kronos_online_scoring.py
abs51295: It's okay if you don't like something, but I would change it :)

util/analytics_platform_util.py
abs51295: Sure.

tests/unit_tests/test_kronos_pgm_pomegranate.py
abs51295: Yes, but since in this file `assertTrue` was being used everywhere else, I used it for consistency. But I think changing all tests to reflect `assertIsNotNone` would greatly benefit.

rootAvish: So here's the thing, new code should be written the right way, you can fix old code whenever you have the time to do that, otherwise you just end up with one more thing you need to fix.

analytics_platform/kronos/src/kronos_online_scoring.py
rootAvish: What I meant was change that to this.

analytics_platform/kronos/src/kronos_online_scoring.py
rootAvish: Python is generally not written like C++/Java.

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
rootAvish: Performance testing is always a one time activity.

rootAvish: And it's done now.

analytics_platform/kronos/deployment/rest_api.py
rootAvish: So why not pass it around instead of storing it in an application wide context?

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
rootAvish: No, using a counter doesn't require using a comprehension.

rootAvish: There's `Counter.update`.

analytics_platform/kronos/src/kronos_online_scoring.py
rootAvish: Comprehensions are "readable" to anybody who knows Python.

analytics_platform/kronos/pgm/src/pgm_pomegranate.py
abs51295: I talked with @miteshvp on this and he agreed to keep the PGM logs for the moment. I will create an issue to remove the logs once it gets stabilized on prod.

analytics_platform/kronos/src/kronos_online_scoring.py
abs51295: Okay, I would replace it with comprehension then.

tests/unit_tests/test_kronos_pgm_pomegranate.py
abs51295: Okay changing and it creating an issue for changing all the occurrences.

analytics_platform/kronos/src/kronos_online_scoring.py
abs51295: That's what I said.

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
abs51295: I would replace it with comprehension since `Comprehensions are "readable" to anybody who knows Python.` :)

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
abs51295: Yes would rename it and update the docstring.

analytics_platform/kronos/deployment/rest_api.py
abs51295: I don't understand what do you mean by that.

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
rootAvish: So this comprehension won't work.

abs51295: This is not the comprehension we would get. We would get this: ` package_to_frequency_counter = Counter(package for package_set in self.manifest_file
                                               for package in package_set)` and this does work :)

rootAvish: Why would you even do that? Counter accepts a list directly.

```ipython
In [4]: a = Counter([1, 2, 3, 4])

In [5]: a
Out[5]: Counter({1: 1, 2: 1, 3: 1, 4: 1})
```

You don't need to pass each package individually.

rootAvish: You just want the overall frequency of each package right?
```python
a = Counter([package_set for package_set in self.manifest_file])
```

abs51295: We have a list of sets and sets are not hashable. So we need to pass each element and not the complete sets.

rootAvish: Comprehensions are only used where a new list creation is required.

rootAvish: Then typecast it to a list when adding it to counter?

rootAvish:
```ipython

In [8]: for i in [set([1, 2, 3]), set([2, 3, 4]), set ([3, 4, 5])]:
   ...:     a.update(list(i))
   ...:
   ...:

In [9]: a
Out[9]: Counter({1: 1, 2: 2, 3: 3, 4: 2, 5: 1})

```

rootAvish: The advantage counter offers is that it maintains the "top" elements internally.

abs51295: I am creating a Counter object and not a list. It's basically a one liner to count occurrences of every package inside the list of package sets.

abs51295: Can you look at the updated code?

analytics_platform/kronos/deployment/rest_api.py
rootAvish: There's `from flask import g` and then there's `app = Flask(__name__)`. Why is it not a part of the former.

analytics_platform/kronos/apollo/src/apollo_generate_frequency_dict.py
```diff
@@ -0,0 +1,42 @@
...
+                                          additional_path=additional_path)
+        return cls(manifest_file=manifest_file)
+
+    def generate_and_save_frequency_dict(self, output_data_store, additional_path):
+        """Generates and saves frequency dictionary from the manifest_file.
+
+        :param output_data_store: Data store to write the frequency dict to.
+        :param additional_path: Path to store the frequency dict inside the output_data_store.
+        """
+        package_to_frequency_counter = Counter(package for package_set in self.manifest_file
```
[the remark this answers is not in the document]
abs51295: I checked with the code you gave and as a matter of fact you don't even need to convert set into a list while using the `counter.update` method. Also, I did some benchmarking using both the approaches and I got the following results:

- Using current approach: 0.0014841556549072266 seconds
- Using proposed approach: 0.0032978057861328125 seconds

As seen, the current approach is still faster than the proposed one. Also, regarding the "readability" problem I don't see any concern since I used the same nested comprehension 
`manifest_topics = {package_topics for package in manifest for package_topics in
                       package_topic_dict.get(package, [])}` as suggested by you for more "readibilty". If we are okay to use nested set comprehension then I don't think there should be a problem with this.

sara-02:
@abs51295 Thanks.

PR #172, opened 2018-01-09
[WIP] - Experiment caching

miteshvp:


miteshvp:
[test]

miteshvp:
[test]

analytics_platform/kronos/deployment/rest_api.py
```diff
@@ -116,21 +121,32 @@ def train_and_save_kronos():
 def predict_and_score():
     input_json = request.get_json()
 
-    app.logger.info("Analyzing the given EPV")
-    app.logger.info(input_json)
+    # Get the response if already cached earlier
+    # Sort package_list elements first
+    for i in input_json:
+        i['package_list'] = sorted(i['package_list'])
```
rootAvish: Hey @miteshvp I don't think this is going to work the way you want it to. This does not alter the original list.

miteshvp: @rootAvish -  I did check on some local json and it works as expected. Also, the tests does prove the point that caching is in play. Still, what makes you think so?

rootAvish: Oh right sorry, I was thinking along the lines of a normal list. Cool, thanks.

rootAvish:
Are we still not required to merge this?

miteshvp:
@rootAvish - we already did https://github.com/fabric8-analytics/fabric8-analytics-stack-analysis/blob/master/analytics_platform/kronos/deployment/rest_api.py#L138. 
But this is still open if we want to control cache-hit ratio for future use

PR #173, opened 2018-01-11
Add template for local deployment

sara-02:


sara-02:
[test]

msrb:
@CermakM PTAL :)

docker-compose-katappa.template
```diff
@@ -0,0 +1,23 @@
...
+services:
+  kronos:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: kronos:latest
+    network_mode: bridge
+    entrypoint:
+      - /bin/entrypoint.sh
+    environment:
```
msrb: Wouldn't be better to use "env file" here? https://docs.docker.com/compose/environment-variables/#the-env-file

sara-02: @msrb we already have that in our backlog to use env instead

sara-02:
@CermakM feel free to send a PR to use env intead ;)

README.md
```diff
@@ -9,6 +9,15 @@
 * [Apollo](/analytics_platform/kronos/apollo)
 * [Uranus](/analytics_platform/kronos/uranus)
 
+## To Deploy Locally:
+
+```bash
+cp docker-compose-katappa.template docker-compose-katappa.yml
```
CermakM: Since this should be part of master branch, I suggest not to use this approach to modify any of native files present in the upstream directory in this way. See comments below.

docker-compose-katappa.template
CermakM: @sara-02 @msrb I do believe env file usage would be much more suitable for this purpose, as it would eliminate the renaming issue, as well as it would be more compact for a potential contributor.

docker-compose-katappa.template
```diff
@@ -0,0 +1,23 @@
+version: "2"
```
CermakM: @sara-02 Any specific reason to use the version "2" instead of "3"?

.gitignore
```diff
@@ -3,7 +3,7 @@ env/
 venv/
 cache/
 config.py
-docker-compose*
+docker-compose*.yml
```
CermakM: I believe we should use *.yml instead of the template, since this would be part of master branch and it may change during development (i.e adding more environment variables). See comments below.

README.md
```diff
@@ -9,6 +9,15 @@
...
+#Add the necessary key-values in docker-compose.yml.
+docker build -t kronos -f Dockerfile
```
CermakM: I think `docker-compose build` command should be used instead of simple `docker` command to avoid contributor confusion and preserve version accordance of the command.

sara-02:
@CermakM sure, please update the PR to bring it upto the mark :)

CermakM:
Reference this PR to #174

PR #177, opened 2018-01-26
Fixed docstrings in the util/datastore module

tisnik:


util/data_store/graph_store.py
```diff
@@ -1,22 +1,34 @@
...
-        """TBD"""
+        """List all the files in the Graph database."""
+        # TODO: not implemented
         return None
 
     def write_json_file(self, filename, contents):
-        """Write JSON file into data source"""
+        """Write JSON file into data source."""
+        # TODO: not implemented
+        pass
```
jpopelka: These methods should rather `raise NotImplementedError`, but I understand it's not part of this PR.

tisnik: yeah, I did not want to change the behaviour (well OTOH this function does not seem to be called anyway ;)

PR #190, opened 2018-02-21
Wildfly support-Part1

sara-02:
For issue https://github.com/openshiftio/openshift.io/issues/2090

### DIFF
--- a/cico_run_pydoc.sh
+++ b/cico_run_pydoc.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -ex
+
+prep() {
+    yum -y update
+    yum -y install epel-release
+    yum -y install python34 python34-virtualenv which
+}
+
+prep
+./check-docstyle.sh


openshift/template.yaml
```diff
@@ -149,4 +149,9 @@ parameters:
   required: true
   name: REPLICAS
   value: "1"
-    
+
+- description: User Post filters or not
+  displayName: User Post filters or not
+  required: true
+  name: USE_FILTERS
+  value: "True"
```
miteshvp: You can directly use a boolean value here https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/openshift/template.yaml#L156

util/analytics_platform_util.py
```diff
@@ -62,3 +62,11 @@ def load_package_list(input_data_store, additional_path):
             for package_list in list_of_package_list:
                 all_list_of_package_set.append(set(package_list))
     return all_list_of_package_set
+
+
+def convert_string2bool_env(parameter):
```
miteshvp: Don't require this if you directly use boolean value in openshift-template file

analytics_platform/kronos/deployment/rest_api.py
```diff
@@ -73,7 +73,7 @@ def setup_logging(flask_app):
                                                                 additional_path=KRONOS_MODEL_PATH)
 
     app.scoring_status = True
-    app.use_filters = USE_FILTERS
+    app.use_filters = convert_string2bool_env(USE_FILTERS)
```
miteshvp: Directly use the boolean value here

openshift/template.yaml
sara-02: @miteshvp In the worker repo the `test` parameter is not getting fetched using the `os.getenv()` function. When we use that function, it returns a string. So even if I pass a boolean value here, when getting fetched using the `os.getenv` it will be converted into string.

PR #201, opened 2018-03-13
Retrain Maven kronos with reduced SLAs

sara-02:
Max SLA observed for package list with length of 10-20 is around 9.5-10.5 ms. Avg around 8.5-9ms

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,11 @@ BODY: JSON data
 ## Latest Deployment
 
 * Maven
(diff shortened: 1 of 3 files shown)


sara-02:
@rootAvish When I tried to test the model with `180/5` `for fp_min_support_count` and `fp_intent_topic_count_threshold` respectively, the scoring time went up for some test cases and decreased for other. Compared to `300/2` where there was a consistent decrease. So we are good with this PR.

Issue #204, opened 2018-04-04
Verify springboot RHOAR alternate recommendation.


sara-02:
Verify the below recommendation:
![screenshot from 2018-04-04 19-40-18](https://user-images.githubusercontent.com/7047079/38312752-28b037ba-3840-11e8-8955-01b423b5fdd5.png)

The tags are a 100% match, and alternate logic is outside of PGM(purely tag-based), so the only way to verify is to check what these packages do.
Booster: `springboot-healthcheck` RHOAR version.

sara-02:
`Springboot-Actuator` is a package to help monitor and manage the health of Springboot applications.
Check this blog for more details about it: https://dzone.com/articles/spring-boot-actuator-a-complete-guide
So expected tags for `springboot-actuator` should be `monitor`, `health`, `metric` etc.
As far as `Sprinboot-hateoas` is a package that helps create hypermedia-driven REST web service using Spring Boot.  
Check this blog for more details about the package: https://dzone.com/articles/applying-hateoas-to-a-rest-api-with-spring-boot
The expected tags for `Springboot-hateoas` should be along the lines of `REST`, `hypermedia` etc.
Since the current tags for  both these packages is `Spring`, `Starter` one is coming as an alternate recommendation for the other.

sara-02:
As per the above details, we need to change the tags for these packages, and retrain the model.

sara-02:
![screenshot from 2018-04-12 17-04-11](https://user-images.githubusercontent.com/7047079/38674870-9a809fa4-3e73-11e8-89ea-2b17359129fe.png)

sara-02:
As we can see the current recommendations make more sense.
Note: They are still based on tag similarity.

PR #207, opened 2018-04-11
Update model

sara-02:
Fix issue #204


sara-02:
Merging this as I have to test on prod-preview

PR #219, opened 2018-05-29
Introduce OpenShift template with restart policy set to Never

msrb:
This new template is for development purposes only. It deploys the
project as a Pod so restart policy can be set to Never. Standard
deployment configs only allow "Always" restart policy.

There is also `template-prod.yaml` which is intended for
staging/production deployments. saas-herder automatically sets restart
policy to "Always" for those environments.

### DIFF
--- a/openshift/deploy.sh
+++ b/openshift/deploy.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/bash -e
+
+set -x
+oc whoami
+oc project
+set +x
+
+function oc_process_apply() {
+  echo -e "\n Processing template - $1 ($2) \n"
+  oc process -f $1 $2 | oc apply -f -
+}
+
+here=`dirname $0`
+template="${here}/template.yaml"
+
+oc_process_apply "$template" "-p KRONOS_SCORING_REGION=maven"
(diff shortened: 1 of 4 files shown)


msrb:
Please review, but do not merge :) I will need to merge https://github.com/openshiftio/saas-analytics/pull/380 together with this PR.

msrb:
cc @sivaavkd

This should do the trick and prevent kronos from crash-looping in dev cluster deployments. It's bit more complicated than I thought it will be.

openshift/deploy.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash -e
```
sara-02: One doubt, we will use this deploy.sh for deploying to dev cluster or at stage-prod?

msrb: To dev cluster, optionally. `deploy.sh` script from https://github.com/fabric8-analytics/fabric8-analytics-deployment/tree/master/openshift will also deploy this service for you (together with all other f8a services). The idea here is that you can use this script to revive your dev cluster kronos deployment, if it failed from some reason.

msrb:
@sara-02 are you OK with me merging this PR?

sara-02:
@msrb Yes. Looks fine to me.

msrb:
Merging. Will also update commit hash in saas-analytics. There are no actual code changes -- this PR is the only change that will be promoted to production.

PR #222, opened 2018-07-10
Move to Quay.io

jmelis:
This PR is part of the effort to move to Quay.

This is the list of changes in this PR:

- Pushes to Quay instead of the Devshift registry
- Uses an image hosted in Quay to build the RHEL based container image
- Uses the new env-toolkit to load variables from jenkins-env
- Changes the default path of the image to quay (note that this should
  not affect production because it is overridden in the saas repo)

A companion PR should have been submitted to the appropriate saas repo
to change the staging url from devshift to quay. The PR to the saas
repo should be merged before this one.
https://github.com/openshiftio/saas-analytics/pull/414

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,6 +1,6 @@
 # base image source https://github.com/rhdt/EL-Dockerfiles/blob/master/base/fabric8-analytics-stack-analysis-base/Dockerfile
 
-FROM prod.registry.devshift.net/osio-prod/base/fabric8-analytics-stack-analysis-base:latest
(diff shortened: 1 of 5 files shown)


msrb:
@sara-02 I'd try to keep this migration to Quay as simple as possible. It's not just image names that often mention "bayesian", or "f8a", or I believe I've seen just "f8" as well. We also use the same names for services, pods, etc. - so fixing just the image names would solve this inconsistency problem only partially.

But yeah, once we have some spare time, we should fix the naming :+1:

jmelis:
@sara-02: I agree with @msrb. It's important to fix this inconsistency, but I would do it outside of the scope of migrating to quay.

Thanks for reviewing :)

